PR #16, opened 2017-09-28
Add -init to create a sample magefile.go

### LINKED ISSUE #8: mage -init
create a new flag -init that creates a new default magefile.

interlock:
* Added -init flag
* Altered magefile() to return on zero files found so we can toggle creating when -init flag set
* Template for magefile.go added
* Generates a magefile.go in the CWD

closes #8


natefinch:
I think this needs to get updated to show the use of deps and a default target, perhaps with those lines commented out and with explanations of what they're for.

interlock:
Added a Default target and a mention of the feature in the readme. The Deps usage was already there, were you thinking something different?

mage/template.go
```diff
@@ -108,3 +108,48 @@ func list() error {
...
+	mg.Deps(Build)
+	fmt.Println("Installing...")
+	cmd := exec.Command("echo", "Pretend to copy someplace else")
+	return cmd.Run()
+}
+
+// Manage your deps, or running package managers.
+func InstallDeps() error {
+	fmt.Println("Installing Deps...")
+	cmd := exec.Command("go", "get", "github.com/magefile/mage")
```
natefinch: can we use a different example dependency?  I don't want people to think they need to go get mage in their magefile.  I know it's dumb, but I also know there's gonna be someone out there that thinks they need this line for their magefile to work.

mage/template.go
```diff
@@ -108,3 +108,48 @@ func list() error {
...
+	return cmd.Run()
+}
+
+// Clean up after yourself
+func Clean() {
+	fmt.Println("Cleaning...")
+	cmd := exec.Command("rm", "MyApp")
```
natefinch: Can you make this os.RemoveAll("MyApp"), to show that it's not just running commands, but also plain old go code?

mage/template.go
```diff
@@ -108,3 +108,48 @@ func list() error {
...
 
 `
+
+var mageTpl = `// +build mage
+
+package main
+
+import (
+	"os/exec"
+	"fmt"
```
natefinch: Super nitpicky, and I'm sorry, but can you ensure this file is go-formatted?  The imports should be sorted, for example.  I know it's a pain, but I don't want the first thing that happens when someone opens the initial magefile is for their editor to reformat it.

mage/template.go
```diff
@@ -108,3 +108,48 @@ func list() error {
...
+	"github.com/magefile/mage/mg" // mg contains helpful utility functions, like Deps
+)
+
+// Default target to run when none is specified
```
natefinch: Specify here also that if no default is set, running just `mage` will list the available targets.

mage/template.go
```diff
@@ -108,3 +108,48 @@ func list() error {
...
+	fmt.Println("Building...")
+	cmd := exec.Command("go", "build", "-o", "MyApp", ".")
+	return cmd.Run()
+}
+
+// A custom install step if you need your bin someplace other than go/bin
+func Install() error {
+	mg.Deps(Build)
+	fmt.Println("Installing...")
+	cmd := exec.Command("echo", "Pretend to copy someplace else")
```
natefinch: can you make this os.Rename("./MyApp",  "/usr/bin/MyApp") or something? It's nice to show that you don't have to drop down to running external commands to get stuff done.

natefinch:
Looks good, but one last thing I missed earlier, please add the new flag to the Usage section of the readme (I believe they're printed in alphabetical order btw).

PR #17, opened 2017-09-29
Fix mg package not getting compiled

interlock:
Package checking only uses .a files and not source files, so we need to compile package mg when building mage. 

I used an underscore import in the mage main to achive this.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,9 @@
 package main
 
-import "github.com/magefile/mage/mage"
+import (
+	"github.com/magefile/mage/mage"
+	_ "github.com/magefile/mage/mg"
+)
 
 func main() {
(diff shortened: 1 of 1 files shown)


natefinch:
That's not really a generic fix.  Any other new import you include in your magefile will cause the same problem.  This is really just a side effect, and it wouldn't work if, for example, you downloaded the mage binary.  We need a better importer that looks in the right places.

interlock:
Agreed. Perhaps a temporary fix. It looks like the go linter tools may be an okay starting point. 

Thoughts on merging this so people can at least get started while we look for a more permanent solution?

natefinch:
this was fixed by #15

PR #23, opened 2017-09-30
Added -keep flag for mage intermediate file

### LINKED ISSUE #14: add --keep flag to prevent generated mainfile from getting deleted
this will help debug problems during generation

interlock:
Added --keep that skips removing the intermediate mage file. 

Readme usage block updated.

closes #14

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -155,7 +155,7 @@ func compile(out string, info *parse.PkgInfo, gofiles []string) error {
 	if err != nil {
 		return errors.WithMessage(err, "can't create mainfile")
 	}
-	if keep == false {
+	if !keep {
 		defer os.Remove(mainfile)
 	}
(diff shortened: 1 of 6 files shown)


mage/main.go
```diff
@@ -154,7 +155,9 @@ func compile(out string, info *parse.PkgInfo, gofiles []string) error {
 	if err != nil {
 		return errors.WithMessage(err, "can't create mainfile")
 	}
-	defer os.Remove(mainfile)
+	if keep == false {
```
natefinch: can you just make this `if !keep {` ?

Also please add a test (I know it's trivial, but I'm trying to make sure new code has tests, and that way we'll know if we break it by accident later).  Tests should go in the mage subdir, since it already has code to set the binary cache to a temp directory that gets cleaned up automatically.

PR #26, opened 2017-10-02
Helpers galore

natefinch:
This commit adds helper functions that should be commonly needed
by many magefiles. It includes command execution handlers as well
as error handlers.

This commit also removes the only two non-stdlib dependencies in
the codebase.  Since mage will likely be vendored into a repo,
it would be nice to avoid cluttering the vendor directory with 
dependencies that are only for the build script.


mage/main.go
```diff
@@ -255,7 +237,7 @@ func run(cmd string, args ...string) int {
 	c.Stdout = os.Stdout
 	c.Env = os.Environ()
 	if verbose {
-		c.Env = append(c.Env, "MAGEFILE_VERBOSE=1")
+		c.Env = append(c.Env, "MAGEFILE_VERVOSE=1")
```
interlock: Typo introduced here, VERBOSE to VER_V_OSE?

natefinch: Yep, thanks!  I added a test for that, too, so next time if I screw it up, at least a test will fail :)

natefinch:
This should land soon. Just testing out the API surface by converting some makefiles to mage.

PR #31, opened 2017-10-03
prepend error output with 'Error:'

### LINKED ISSUE #27: Feature Request : Prefix returned errors with something such as `Error:`
In the video demo your example of `return errors.New("boo!")` just prints out "boo!" and whilst the `echo $?` shows a non-zero exit code, there is nothing on the screen to visually show the command failed.

I know some people have prompts which display something when the previous command failed, however I think it would be nice to show it visually too. Nothing too big, but just big enough.

For example:

```
$ mage install
Error: boo!
```

As a comparison let's just look at what `make` does. We don't have to copy this nor am I holding it up as the ideal, but just a curiosity. For example, if my target just tried to cat a non-existant file:

```
$ make target
cat hello.go
cat: hello.go: No such file or directory
Makefile:2: recipe for target 'target' failed
make: *** [build] Error 1
```

Which (I think) is made up of the command, stderr from the command and then output from make itself. I'm happy to keep it real simple as above and just prefix the output with `Error:`.

(Note: I think we could or should also do the same for functions that don't return errors, so I'm not sure how we'd do that - perhaps if they wrote something on stderr? This comment is a non-goal of this ticket and perhaps another ticket should be opened, however it's pertinent to raise this here in case it impacts what we do here.)

chilts:
fixes #27

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -46,7 +46,7 @@ func main() {
 	defer func() {
 		err := recover()
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("Error: %v\n", err)
 			type code interface { ExitStatus() int }
 			if c, ok := err.(code); ok {
 				os.Exit(c.ExitStatus())
@@ -79,7 +79,7 @@ func main() {
 	case "{{lower .Name}}":
 		{{if .IsError -}}
 			if err := {{.Name}}(); err != nil {
-				fmt.Println(err)
+				fmt.Printf("Error: %v\n", err)
 				os.Exit(1)
 			}
 		{{else -}}
--- a/mage/main_test.go
+++ b/mage/main_test.go
@@ -75,6 +75,57 @@ func TestVerbose(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


mage/template.go
```diff
@@ -46,7 +46,7 @@ func main() {
 	defer func() {
 		err := recover()
 		if err != nil {
-			fmt.Println(err)
+			fmt.Printf("Error: %s\n", err)
```
natefinch: this and the one below need to be %v in case someone panics with something that isn't compatible with %s.

chilts:
Force pushed with those two as %v instead of %s. Thanks for the tip.

natefinch:
looks great, thanks!

chilts:
Not sure if there is anything I can do about that netlify deploy. I'm guessing that's what is blocking this rather than the CI tests, which look like they passed ok.

natefinch:
yep, don't worry about netlify. that's just because there's no actual site to deploy yet :)  but there will be soon, so I turned it on to make sure it works.

PR #32, opened 2017-10-03
Feature/context for jobs

### LINKED ISSUE #25: Feature Request: automatic insertion of a job-scoped context
This would allow for users to write functions like:

```go
func Build(ctx context.Context) error {
  cmd := exec.CommandContext(ctx, "/usr/local/go/bin/go", "build", "./cmd/...")
  cmd.Stdout = os.Stdout
  cmd.Stderr = os.Stderr
  err := cmd.Run()
}
```

And then if the user hits control-c to terminate the mage process, the context will instantly cancel and cause the process started by the `Build` function to be killed.

Mage itself would need to start and manage these contexts.

It would be nice to also have a command line flag `-t` that would set a timeout on these contexts.

interlock:
Allows for jobs that take and call with a context as described in #25

Todo:
- [x] Context caller for Deps as a separate func helper
- [x] Test a simple Timeout context
- [x] split dup code for acceptable func types out
- [x] wrap calling job func's to deal with error/context uniformly

closes #25


natefinch:
gah, sorry for the conflicts.

interlock:
Ya, I saw it coming from that other PR. Still working out handling timeouts locally.

example/magefile.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+)
+
+// Default target to run when none is specified
+// If not set, running mage will list available targets
+// var Default = Build
+
+func Timeout(ctx context.Context) {
+	fmt.Printf("timeout start\n")
+	time.Sleep(1000000000000)
```
natefinch: time.sleep takes a time.Duration, you could make this easier to read by just doing time.Sleep(time.Hour) or something :)

mage/main.go
```diff
@@ -35,8 +35,8 @@ const initFile = "magefile.go"
 
 var (
 	force, verbose, list, help, mageInit, keep, showVersion bool
-
-	timestamp, commitHash, gitTag string
+	timeout                                                 int
```
natefinch: you can put spaces between the lines os the types aren't so crazily right justified :)

example/magefile.go
interlock: It does and "oops!" did not intend to commit that... I'll remove it. I use that for my manual test cases.

interlock:
This goes as far as I can take it right now.
- Parse updated to spit out some scaffold around running targets wrapped with context
- template has a new function, `runTarget` that handles the defer recover and actual ctx channel handling for errors. Brings the extraction of a StatusCode error to one spot as well.
- There is a single test for expiring a context on timeout. We could have more complicated tests, but I figure we can make those when someone does something complicated.
- I lightly updated the site docs to indicate there is now a `-t <seconds>` flag to set a timeout
- `mg/runtime` has a `Context` function which returns the same Context everytime based on if we have a `-t` or not.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+// Place your settings in this file to overwrite default and user settings.
+{
```
natefinch: this should probably be .gitignored

mage/main.go
```diff
@@ -40,6 +40,8 @@ const mainfile = "mage_output_file.go"
 const initFile = "magefile.go"
 
 var (
+	force, verbose, list, help, mageInit, keep, showVersion bool
```
natefinch: this can get removed.... with switch to a standalone parse function, we don't have global flag variables anymore.

mage/main_test.go
```diff
@@ -31,7 +31,11 @@ func testmain(m *testing.M) int {
 		log.Fatal(err)
 	}
 	if err := os.Mkdir(dir, 0700); err != nil {
-		log.Fatal(err)
+		if os.IsExist(err) {
+			os.RemoveAll(fmt.Sprintf("%s/*", dir))
```
natefinch: RemoveAll removes a directory.  globbing (*) doesn't generally work in go functions anyway.  if you want to just delete everything *inside* the directory, you'll either have to loop through, or better, just delete the whole thing and recreate it.

mage/main.go
```diff
@@ -58,6 +60,7 @@ type Invocation struct {
 	List    bool      // tells the magefile to print out a list of targets
 	Help    bool      // tells the magefile to print out help for a specific target
 	Keep    bool      // tells mage to keep the generated main file after compiling
+	Timeout int       // tells mage to set a timeout to running the targets
```
natefinch: make this a time.Duration, please.

mage/main.go
```diff
@@ -115,6 +118,7 @@ func Parse(stdout io.Writer, args []string) (inv Invocation, mageInit, showVersi
 	fs.BoolVar(&inv.List, "l", false, "list mage targets in this directory")
 	fs.BoolVar(&inv.Help, "h", false, "show this help")
 	fs.BoolVar(&mageInit, "init", false, "create a starting template if no mage files exist")
+	fs.IntVar(&inv.Timeout, "t", 0, "timeout in seconds, Default (0) no timeout")
```
natefinch: I think this is better as a DurationVar... it's nice to be able to do `-t 30s` or `-t 5m` etc

Even if we decide to keep the flag as an int, the Invocation Timeout field should definitely be a time.Duration, it makes testing easier, and it's the correct type for durations :)

mage/template.go
```diff
@@ -6,12 +6,17 @@ var tpl = `// +build ignore
...
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
+
 	"strings"
+
```
natefinch: it doesn't really matter, but it would be nice to clean up these extra blank lines

mg/deps.go
```diff
@@ -95,18 +103,25 @@ func changeExit(old, new int) int {
...
-		fn = func() error { f(); return nil }
+		fn = func(ctx context.Context) error { f(); return nil }
 	case func() error:
+		fn = func(ctx context.Context) error { return f() }
+	case func(context.Context):
+		fn = func(ctx context.Context) error { f(ctx); return nil }
+	case func(context.Context) error:
 		fn = f
+	default:
+		panic("attempted to add a dep that did not match required type")
```
natefinch: can you slap a 
`// should be impossible, since we already checked this`
above this?

mg/runtime.go
```diff
@@ -34,3 +38,26 @@ func CacheDir() string {
 		return filepath.Join(os.Getenv("HOME"), ".magefile")
 	}
 }
+
+var ctx context.Context
+var ctxCancel func()
+
+func GetContext() (context.Context, func()) {
```
natefinch: Can you call this just Context()?  GetFoo is sorta non-idiomatic.

mg/runtime.go
```diff
@@ -34,3 +38,26 @@ func CacheDir() string {
...
+	if ctx != nil {
+		return ctx, ctxCancel
+	}
+
+	if os.Getenv("MAGEFILE_TIMEOUT") != "" {
+		timeout, err := strconv.Atoi(os.Getenv("MAGEFILE_TIMEOUT"))
+		if err != nil {
+			fmt.Println("timeout must be a number >= 0")
```
natefinch: speaking of which, check for negative numbers here

parse/parse.go
```diff
@@ -10,19 +10,64 @@ import (
...
-	Synopsis string
-	Comment  string
+	Name      string
+	IsError   bool
+	IsContext bool
+	Synopsis  string
+	Comment   string
+}
+
+func (f Function) TemplateString() string {
```
natefinch: doc comment here please

parse/parse_test.go
```diff
@@ -11,24 +12,48 @@ func TestParse(t *testing.T) {
...
+		},
+	}
+
+	// DefaultIsError
+	if info.DefaultIsError != true {
+		t.Fatalf("expected DefaultIsError to be true")
+	}
+
+	// DefaultName
+	if strings.Compare(info.DefaultName, "ReturnsError") != 0 {
```
natefinch: `if info.DefaultName != "ReturnsError" {`

mage/main.go
```diff
@@ -368,7 +367,7 @@ func RunCompiled(inv Invocation, exePath string) int {
 		c.Env = append(c.Env, "MAGEFILE_HELP=1")
 	}
 	if inv.Timeout > 0 {
-		c.Env = append(c.Env, fmt.Sprintf("MAGEFILE_TIMEOUT=%d", inv.Timeout))
+		c.Env = append(c.Env, fmt.Sprintf("MAGEFILE_TIMEOUT=%dns", inv.Timeout))
```
natefinch: you can probably just do inv.Timeout.String()

mage/main_test.go
```diff
@@ -32,7 +33,7 @@ func testmain(m *testing.M) int {
 	}
 	if err := os.Mkdir(dir, 0700); err != nil {
 		if os.IsExist(err) {
-			os.RemoveAll(fmt.Sprintf("%s/*", dir))
+			os.RemoveAll(fmt.Sprintf("%s", dir))
```
natefinch: can remove the sprintf now :)

PR #35, opened 2017-10-03
Make demo thumbnail hi-res

sparrc:


natefinch:
Much better, thank you :)

PR #39, opened 2017-10-04
present the commands in camelCase - fixes #37

### LINKED ISSUE #37: Camel case the commands in listing
I get that this is very subjective, but this is from the Hugo build:

```bash
▶ mage
  check            Run tests and linters
  checkvendor      Verify that vendored packages match git HEAD
  docker           Build hugo Docker container
  fmt              Run gofmt linter
  hugo             Build hugo binary
  hugonogitinfo    Build hugo without git info
  hugorace         Build hugo binary with race detector enabled
  install          Install hugo binary
  lint             Run golint linter
  test             Run tests
  test386          Run tests in 32-bit mode
  testcoverhtml    Generate test coverage report
  testrace         Run tests with race detector
  vendor           Install govendor and sync Hugo's vendored dependencies
  vet              Run go vet linter
```

Which is mostly beautiful. But the all lower-case command names looks "wrong" in my eyes. I would much prefer that it followed the original case of the method, but with a lowercase first letter, so `hugoRace`, `testCoverHTML`etc. I would still expect the all lowercase variant to work.

chilts:


chilts:
I added a new test case to cover the listing too. Currently we don't test the listing anywhere (I know this because with my code change, the tests still passed.)

The new test case files are needed because currently when we do a `TestGoRun` test, they all have `Default` targets and so the list isn't done because the Default exists. The new test has no default, hence the list is shown.

natefinch:
so fyi, you can do `go run main.go -l` and it'll do the list, but it's useful to have a test for a no-default setup anyway.

mage/main.go
```diff
@@ -27,6 +27,9 @@ const mageVer = "v0.3"
 
 var output = template.Must(template.New("").Funcs(map[string]interface{}{
 	"lower": strings.ToLower,
+	"lowerfirst": func(s string) string {
+		return strings.ToLower(string(s[0])) + s[1:]
```
natefinch: this is probably fine 99% of the time, but it technically can break with unicode characters that are more than a single byte, since slicing a string returns bytes, not unicode codepoints (runes).  

A safer (if less elegant) way to do it is 

```go
r := []rune(s)
r = append([]rune{unicode.ToLower(r[0])}, r[1:])
return string(r)
```

mage/main_test.go
```diff
@@ -75,6 +75,21 @@ func TestVerbose(t *testing.T) {
...
+func TestGoRunList(t *testing.T) {
+	c := exec.Command("go", "run", "main.go")
+	c.Dir = "./testdata/no_default"
+	c.Env = os.Environ()
+	b, err := c.CombinedOutput()
+	if err != nil {
+		t.Error("error:", err)
+	}
+	actual := string(b)
+	expected := "  bazBuz    Prints out 'BazBuz'.\n  fooBar    Prints out 'FooBar'.\n"
```
natefinch: this is nitpicky, but can you make this a multiline string, i.e. use \`\` instead?  I think it'll read a lot easier.  A trick I use is to use a line return at the beginning of the string and then immediately slice it off, to make it easy to read, e.g.  

```
    expected := `
  bazBuz     Prints out 'BazBuz'.
  fooBar .   Prints out 'FooBar'.
`[1:]

The slice then removes that initial line return, but it keeps the string all nicely left adjusted.

mage/main.go
chilts: I did stumble across a rune example doing this, though I failed to spot that the slice would break with unicode chars. Thanks for noting.

mage/main_test.go
chilts: Perfect. Will do.

chilts:
Updated PR. Just a few notes:

1. it didn't quite like your append with the `r[1:]` (https://play.golang.org/p/8bxZdnFw03 - `cannot use r[1:] (type []rune) as type rune in append`) so I did it slightly differently.
2. I added a test for the `-l` flag too. The only thing I noticed here is that when printing out the targets with just `mage` compared to `mage -l` the latter also adds a first line of "Targets:<space>" whereas the former doesn't. It's a minor inconsistency.
3. I had to start the multiline string straight after the opening backtick due to the initial newline.
4. When printing out a target with no description, we get "<space><space>targetname<lots-of-spaces>" which is again minor but just means the expected test strings have spaces at the end of some lines. Again, not an issue, just a note.

natefinch:
oh yeah, I forgot the ... after the second.

so should have been

`append([]rune{unicode.ToLower(r[0])}, r[1:]...)`

and actually, I'm silly, because this would be even simpler:

r := []rune(s)
r[0] = unicode.ToLower(r[0])

Duh.. but anyway, this still works just fine.  Thanks!

Issue #30, opened 2017-10-03
Add MAGE_VERBOSE flag to permanently turn on verbose mode


natefinch:
I basically always want to run with verbose mode on, so it would be nice if I could just set an environment variable and not need to remember to type -v.

natefinch:
Done! thanks @webRat

PR #43, opened 2017-10-05
Issue #30 - Added env check "MAGE_VERBOSE" to mage.

webRat:


webRat:
Actually hold off and I'll see if I can add a test too. :o

Related to Issue #43

natefinch:
can you check for 0/1 as well?  I would like MAGE_VERBOSE=1 to do the right thing.  Those monsters who use MAGE_VERBOSE=T can get stuffed, though.

webRat:
I think I'm going to have to start over. The main.go file completely changed on me.

(╯°□°)╯︵ ┻━┻

natefinch:
the new changes should make this a lot easier.  The logic is a lot more broken up into individual functions that can be unit tested.  So just throw this code in Parse... it'll make your life a lot easier.

webRat:
🙌

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Launch",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
+            "remotePath": "",
+            "port": 2345,
```
natefinch: can you add .vscode to the .gitignore?

mage/main.go
```diff
@@ -137,6 +138,15 @@ func Parse(stdout io.Writer, args []string) (inv Invocation, mageInit, showVersi
 func Invoke(inv Invocation) int {
 	log := log.New(inv.Stderr, "", 0)
 
+	// If verbose is still false, we're going to peek at the environment variable to see if
+	// MAGE_VERBOSE has been set. If so, we're going to use it for the value of MAGE_VERBOSE.
+	if inv.Verbose == false {
```
natefinch: this feels like it belongs in Parse().  Parse is where we take environmental factors (like os.Args and os.Environ) and convert them into function arguments (in our case, an Invocation struct)

natefinch: Ideally I'd like ParseAndRun to take a []string{} of the same type os.Environ returns, and then pass that to Parse.  Then Main() can just pass os.Environ() ... and it makes testing easy without worring about environment variables on the test machine.

natefinch: And then you could just test Parse and not need to worry about actually calling Invoke, since that would just be testing the Verbose field in Invocation, which should be separate.

natefinch: If this sounds like too much work, feel free to just pop this into parse, and I'll do the rest :)

webRat: Originally I had it in parse.Test was failing because the Test was calling 'Invoke()' directly. if it's in Parse, Invoke will never see it.

webRat: Picking up my kids. Will peek at this again when I get back.

PR #45, opened 2017-10-05
Prevent multiple targets

### LINKED ISSUE #42: Either fail or support multiple targets
See https://github.com/gohugoio/hugo/pull/3939

Good thing I tested.

So I just replaced `make` with `mage` and got:

```bash
mage hugoRace check
```

Wich does not fail. If you do not support multiple targets you really should exit with error + some error message.

natefinch:
We don't support multiple targets, so let's fail if people
try to use more than one.  This could change in the future,
but right now, extra args are silently ignored which is bad
when you think silence means that second target succeeded.

This fixes #42.


natefinch:
This also fixes a bug I just created with my last PR, where everything was just printing out help.  Also adds a test for both these changes.

PR #49, opened 2017-10-05
Fix broken godoc link in README

y0ssar1an:
`godoc.org/github.com/mage/mage` -> `godoc.org/github.com/magefile/mage/mage`


natefinch:
oops, thanks! :)

PR #52, opened 2017-10-07
fix homepage

natefinch:


natefinch:
no longer needed , just needed newer version on hugo on netlify

PR #54, opened 2017-10-09
Add mg.Target

gernest:
This function allows specifying files/directories as targets and returns true when the targets are met and false otherwise.

Targets are met if the ModTime of src is older than any ModTime of any  of the specified targets.

For recursive checks of directory with its file contents set recusive argument to true

### DIFF
--- a/mg/target.go
+++ b/mg/target.go
@@ -1,0 +1,92 @@
+package mg
+
+import (
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// Target checks whether the target files/directories have been more recently
+// modified than src.
+//
+// When recursive is set to true, if src is a directory it will be walked
+// recursively to deermine modified time.
+func Target(src string, recursive bool, targets ...string) (bool, error) {
+	if len(targets) == 0 {
+		return true, nil
+	}
+	stat, err := os.Stat(src)
+	if err != nil {
+		return false, err
+	}
+	srcTime := stat.ModTime()
+	if stat.IsDir() && recursive {
+		srcTime = calDirModTimeRecursive(stat)
+	}
+	dt, err := loadTargets(targets)
+	if err != nil {
+		return false, err
+	}
+	t := dt.modTime(false)
+	if t.After(srcTime) {
+		return true, nil
+	}
+	if dt.hasdir {
+		return dt.modTime(true).After(srcTime), nil
+	}
+	return false, nil
+}
+
+func calDirModTimeRecursive(dir os.FileInfo) time.Time {
+	t := dir.ModTime()
+	filepath.Walk(dir.Name(), func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.ModTime().After(t) {
+			t = info.ModTime()
+		}
+		return nil
+	})
+	return t
+}
+
+type depTargets struct {
+	src    []os.FileInfo
+	hasdir bool
+	latest time.Time
+}
+
+func loadTargets(targets []string) (*depTargets, error) {
+	d := &depTargets{}
+	for _, v := range targets {
+		stat, err := os.Stat(v)
+		if err != nil {
+			return nil, err
+		}
+		if stat.IsDir() {
+			d.hasdir = true
+		}
+		d.src = append(d.src, stat)
+		if stat.ModTime().After(d.latest) {
+			d.latest = stat.ModTime()
+		}
+	}
+	return d, nil
+}
+
+func (d *depTargets) modTime(recursive bool) time.Time {
+	if !recursive || !d.hasdir {
+		return d.latest
+	}
+	for _, i := range d.src {
+		t := i.ModTime()
+		if i.IsDir() {
+			t = calDirModTimeRecursive(i)
+		}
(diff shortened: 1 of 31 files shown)


mg/target.go
```diff
@@ -0,0 +1,92 @@
...
+	"path/filepath"
+	"time"
+)
+
+// Target checks whether the target files/directories have been more recently
+// modified than src.
+//
+// When recursive is set to true, if src is a directory it will be walked
+// recursively to deermine modified time.
+func Target(src string, recursive bool, targets ...string) (bool, error) {
```
platelk: The name could be more self helpful, here `Target` could mean the target of an execution or something else. maybe something like `HasBeenModify` or `IsUpToDate` could be more meaningful

mg/target.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if dt.hasdir {
+		return dt.modTime(true).After(srcTime), nil
+	}
+	return false, nil
+}
+
+func calDirModTimeRecursive(dir os.FileInfo) time.Time {
+	t := dir.ModTime()
+	filepath.Walk(dir.Name(), func(path string, info os.FileInfo, err error) error {
```
natefinch: filepath.Walk can return an error, you need to check it.

mg/target.go
```diff
@@ -0,0 +1,92 @@
...
+	"time"
+)
+
+// Target checks whether the target files/directories have been more recently
+// modified than src.
+//
+// When recursive is set to true, if src is a directory it will be walked
+// recursively to deermine modified time.
+func Target(src string, recursive bool, targets ...string) (bool, error) {
+	if len(targets) == 0 {
```
natefinch: I would panic here... this is a programmer error... there's no reason to call this with less than one prerequisite.

mg/target.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		d.src = append(d.src, stat)
+		if stat.ModTime().After(d.latest) {
+			d.latest = stat.ModTime()
+		}
+	}
+	return d, nil
+}
+
+func (d *depTargets) modTime(recursive bool) time.Time {
```
natefinch: please split this up into two functions, I don't like using booleans to split behavior like this.

mg/target_test.go
```diff
@@ -0,0 +1,64 @@
...
+		"testdata/src_file",
+		"testdata/target_dir/some_file",
+		"testdata/target_file",
+	}
+	for _, v := range files {
+		err := ioutil.WriteFile(v, []byte(v), 0600)
+		if err != nil {
+			t.Fatal(err)
+		}
+		time.Sleep(time.Second)
```
natefinch: you can make this sleep 10ms.... no need to make this test take 5 seconds.  10ms will still be plenty to spread out the file creation times.

mg/target_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	defer func() {
+		err := os.RemoveAll("testdata")
```
natefinch: we may well have a permanent testdata dir in this directory, so can you please move these tests into a subdir of testdata, so that this won't remove permanent files we need at some point? :)

mg/target.go
natefinch: I think this could do well with being moved to a separate package, to help with namespacing.  Maybe call the package target, and then you could have `target.Dir(path string, prerequisites ....string)` which would be recursive and `target.Path(path string, prerequisites ...string)` which would not be recursive.   This is nicer than a boolean, which I dislike using to change logic of a function because it is obscure when you're reading it. `Target("foo", true, "bar", "baz")` <- when reading this, I don't know what "true" means, unless I got look up the definition of the function.

gernest:
@natefinch @platelk  please check this again, any notes are welcome.

target/target.go
```diff
@@ -0,0 +1,108 @@
+package mg
```
natefinch: this should be package target

target/target.go
```diff
@@ -0,0 +1,108 @@
...
+	t := dt.modTimeDir()
+	if t.After(srcTime) {
+		return true, nil
+	}
+	return false, nil
+}
+
+func calDirModTimeRecursive(dir os.FileInfo) time.Time {
+	t := dir.ModTime()
+	filepath.Walk(dir.Name(), func(path string, info os.FileInfo, err error) error {
```
natefinch: filepath.Walk can return an error, so let's not drop it

target/target_test.go
```diff
@@ -0,0 +1,63 @@
...
+		"testdata/src_file",
+		"testdata/target_dir/some_file",
+		"testdata/target_file",
+	}
+	for _, v := range files {
+		err := ioutil.WriteFile(v, []byte(v), 0600)
+		if err != nil {
+			t.Fatal(err)
+		}
+		time.Sleep(time.Second)
```
natefinch: can you make this time.Millisecond*10 or something? it doesn't have to be a second to get different timestamps.

target/target_test.go
```diff
@@ -0,0 +1,63 @@
+package mg
```
natefinch: package target

gernest:
Done

PR #73, opened 2017-10-16
website: typo on dependencies page. g should depend on h, not on itself

jasonhancock:
I believe there's a typo on https://magefile.org/dependencies/. Currently, `g` depends on itself but it looks like it should have depended on `h` instead.


natefinch:
lol, yes, thanks! 

That's not a useful dependency (ironically, it won't infinite loop, since dependencies are guaranteed only to run once... but it's still not a useful dependency, and definitely an anti-pattern.

PR #81, opened 2017-10-23
Cleaned up use of mage/types package

interlock:
* Removed stale type declarations in parse.
* Updated parse and mg to make better use of types package.
* Added function to types package to test for valid function.
* Added function to types package to wrap a valid FuncType as the most greedy of FuncType

### DIFF
--- a/mg/deps.go
+++ b/mg/deps.go
@@ -106,8 +106,8 @@ func runDeps(ctx context.Context, fns ...interface{}) {
 
 func checkFns(fns []interface{}) {
 	for _, f := range fns {
-		if !types.IsFuncType(f) {
-			panic(fmt.Errorf("Invalid type for dependent function: %T. Dependencies must be func(), func() error, func(context.Context) or func(context.Context) error", f))
+		if err := types.FuncCheck(f); err != nil {
+			panic(err)
 		}
 	}
 }
--- a/mg/deps.go
+++ b/mg/deps.go
@@ -7,6 +7,8 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+
+	"github.com/magefile/mage/types"
 )
 
 type onceMap struct {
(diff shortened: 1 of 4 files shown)


mg/deps.go
```diff
@@ -104,10 +106,7 @@ func runDeps(ctx context.Context, fns ...interface{}) {
 
 func checkFns(fns []interface{}) {
 	for _, f := range fns {
-		switch f.(type) {
-		case func(), func() error, func(context.Context), func(context.Context) error:
-			// ok
-		default:
+		if !types.IsFuncType(f) {
```
natefinch: can you make this CheckFunc(f) and having it return an error? That way you wouldn't need to hard code the error text in every place we use this, it would just be part of the types package code.

types/funcs.go
```diff
@@ -27,3 +27,43 @@ type FuncContext func(context.Context)
...
+	case func(context.Context):
+		return true
+	case func(context.Context) error:
+		return true
+	}
+	return false
+}
+
+// FuncTypeWrap wraps a valid FuncType to FuncContextError
+func FuncTypeWrap(fn interface{}) FuncContextError {
```
natefinch: so I don't think the named type FuncContextError really pays for itself here. It doesn't have any special meaning.  I think it's clearer just to make this 
```
func FuncTypeWrap(fn interface{}) func(context.Context) error {
```

Then you don't have to look up what FuncContextError is to figure out what this is returning.  And since the returned function is not special in any way, it's nice not fool people into thinking it's something special by giving it a name.

types/funcs.go
```diff
@@ -28,24 +29,24 @@ type FuncContext func(context.Context)
 // FuncContextError takes a context and may return an error
 type FuncContextError func(context.Context) error
```
natefinch: Can you remove this type now?

PR #84, opened 2017-10-24
Fixed backward compatibility

natefinch:
This PR removes the use of mg.Context from the generated mainfile,
instead generating the code to create the context in the mainfile
itself. This prevents backwards compatibility problems with older
versions of mage that may be vendored, which don't have mg.Context.

Also adds a test to ensure that we don't have the generated mainfile
depend on anything outside the stdlib.

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -40,8 +40,11 @@ var initOutput = template.Must(template.New("").Parse(mageTpl))
 const mainfile = "mage_output_file.go"
 const initFile = "magefile.go"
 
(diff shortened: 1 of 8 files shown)


natefinch:
This fixes https://github.com/gohugoio/hugo/issues/4003

vassudanagunta:
I confirmed that this fixes gohugoio/hugo#4003. Thanks!

natefinch:
Super awesome @vassudanagunta!  So glad to hear it :)

Issue #58, opened 2017-10-11
Support for Aliases


natefinch:
Let users define a list of aliases for targets:

```go
var Aliases = map[string]interface{}{
    "example-foo" : ExampleFoo,
}
```

Alias names would be compared vs. the lowercased target name specified on the command line.  The code would check for alias matches first, and if found, use reflection to get the function name specified, and replace the CLI target name with the reflection generated name.

natefinch:
implemented in #90

PR #90, opened 2017-10-30
Support for aliases (#58)

mbialon:
I've added support for aliases. What do you think about these changes?

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -31,6 +31,8 @@ func main() {
 	}
 
 	targets := map[string]bool {
+		{{range $alias, $funci := .Aliases}}"{{lower $alias}}": true,
+		{{end}}
 		{{range .Funcs}}"{{lower .Name}}": true,
 		{{end}}
 	}
@@ -58,7 +60,15 @@ func main() {
 		switch strings.ToLower(os.Args[1]) {
 			{{range .Funcs}}case "{{lower .Name}}":
 				fmt.Print("mage {{lower .Name}}:\n\n")
-                fmt.Println({{printf "%q" .Comment}})
+				{{if ne .Comment ""}}fmt.Println({{printf "%q" .Comment}}){{end}}
+				var aliases []string
+				{{- $name := .Name -}}
+				{{range $alias, $func := $.Aliases}}
+				{{if eq $name $func}}aliases = append(aliases, "{{$alias}}"){{end -}}
+				{{- end}}
+				if len(aliases) > 0 {
+					fmt.Printf("Aliases: %s\n\n", strings.Join(aliases, ", "))
+				}
 				return
 			{{end}}
 			default:
@@ -82,6 +92,12 @@ func main() {
 	{{- end}}
 	}
 	for _, target := range os.Args[1:] {
+		switch strings.ToLower(target) {
+		{{range $alias, $func := .Aliases}}
+		case "{{lower $alias}}":
+			target = "{{$func}}"
+		{{- end}}
+		}
 		switch strings.ToLower(target) {
 		{{range .Funcs }}
 		case "{{lower .Name}}":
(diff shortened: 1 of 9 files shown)


natefinch:
From a first once-over, this looks awesome.  I'll give it a more detailed review tonight. Thanks for contributing!

parse/parse.go
```diff
@@ -141,6 +143,41 @@ func setDefault(p *doc.Package, pi *PkgInfo, info types.Info) {
 	}
 }
 
+func setAliases(p *doc.Package, pi *PkgInfo, info types.Info) {
+	for _, v := range p.Vars {
+		for x, name := range v.Names {
+			if name != "Aliases" {
+				continue
+			}
+			spec := v.Decl.Specs[x].(*ast.ValueSpec)
```
natefinch: You need to make this `spec, ok :=`  and then check for ok, otherwise the next line will panic if you have a function named Aliases, for example.

parse/parse.go
```diff
@@ -141,6 +143,41 @@ func setDefault(p *doc.Package, pi *PkgInfo, info types.Info) {
...
+			if len(spec.Values) != 1 {
+				log.Println("warning: aliases declaration has multiple values")
+			}
+			comp, ok := spec.Values[0].(*ast.CompositeLit)
+			if !ok {
+				log.Println("warning: aliases declaration is not a map")
```
natefinch: return after this line, since we can't continue.

parse/parse.go
```diff
@@ -141,6 +143,41 @@ func setDefault(p *doc.Package, pi *PkgInfo, info types.Info) {
...
+			}
+			pi.Aliases = make(map[string]string)
+			for _, elem := range comp.Elts {
+				kv, ok := elem.(*ast.KeyValueExpr)
+				if !ok {
+					log.Println("warning: alias declaration is not a map element")
+				}
+				k, ok := kv.Key.(*ast.BasicLit)
+				if !ok || k.Kind != token.STRING {
+					log.Println("warning: alias is not a string")
```
natefinch: return here

parse/parse.go
```diff
@@ -141,6 +143,41 @@ func setDefault(p *doc.Package, pi *PkgInfo, info types.Info) {
...
+				}
+				v, ok := kv.Value.(*ast.Ident)
+				if !ok {
+					log.Println("warning: alias target is not a function")
```
natefinch: return here

mage/template.go
```diff
@@ -31,6 +31,8 @@ func main() {
 	}
 
 	targets := map[string]bool {
+		{{range $alias, $funci := .Aliases}}"{{$alias}}": true,
```
natefinch: you should lowercase the alias, because we always lowercase the input, so an alias of "Foo" will never match.

mage/template.go
```diff
@@ -83,6 +85,12 @@ func main() {
 	}
 	for _, target := range os.Args[1:] {
 		switch strings.ToLower(target) {
+		{{range $alias, $func := .Aliases -}}
+		case "{{$alias}}":
```
natefinch: this should be lowercased, too

parse/parse.go
```diff
@@ -109,6 +110,7 @@ func Package(path string, files []string) (*PkgInfo, error) {
 	}
 
 	setDefault(p, pi, info)
+	setAliases(p, pi, info)
```
natefinch: one more thing, can you sanity check that all the aliases refer to valid targets?   Thanks!

natefinch:
If you'd like to tackle it, it would be nice to add the aliases to the mage -h output for a specific target.

For example, if mage's own magefile had 

```go
var Aliases = map[string]interface{} {
  "i":     Install,
  "build": Install,
}
```

```
$ mage -h install
mage install:

Runs "go install" for mage.  This generates the version info the binary.

Aliases: i, build
```

mbialon:
Thanks for reviewing the changes. I've fixed the issues and I've added a list of aliases to target's help output.

natefinch:
Awesome, thanks, I'll review shortly.

natefinch:
Looks great. Thanks for this! :)

PR #95, opened 2017-12-28
parse: sanitize function doc synopsis

### LINKED ISSUE #50: Remove stutter in `mage -l`
When writing things using function commentary such as:

```go
// Build builds and re-generates all needed dependencies and compiles all binaries for $APP_NAME.
func Build() {
// ... etc
```

This gets recorded in the mage task list as:

```console
$ mage -l 
Targets:
  build      Build builds and re-generates all needed dependencies and compiles all binaries for $APP_NAME.
```

What I would expect would be something akin to:

```console
$ mage -l 
Targets:
  build      builds and re-generates all needed dependencies and compiles all binaries for $APP_NAME.
```

I have a proof of concept patch [here](https://gist.github.com/Xe/9029aa078953b14e026c827765fdba1b) that implements this in a rather hacky way that seems to work pretty well.

darkowlzz:
This change adds `sanitizeSynopsis()` function to perform basic
sanitization on the synopsis. If synopsis begins with the function name, it
removes the function to avoid repeating words.

Sorry if I've added too many comments trying to leave hints for the particular piece of code. I can remove them if they are silly 😅

Fixes #50

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -102,7 +102,7 @@ func Package(path string, files []string) (*PkgInfo, error) {
 			pi.Funcs = append(pi.Funcs, Function{
 				Name:      f.Name,
 				Comment:   f.Doc,
-				Synopsis:  doc.Synopsis(f.Doc),
+				Synopsis:  sanitizeSynopsis(f),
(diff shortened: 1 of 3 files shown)


Xe:
sorry to necro this but 👍

Issue #94, opened 2017-12-22
Make --init and --version into "commands"


natefinch:
bail if either flag is set if fs.NArgs() > 0 || fs.NFlags() > 1 - i.e. if any other args were specified.

Also move them into a separate section of the usage output called "commands".

The --clean flag should be treated the same way from #19

vassudanagunta:
Looks like the functionality has been implemented, since when a commands is given, any targets also specified are not executed. Sweet. But rather than ignoring them silently, it would be better if there was a warning. This should apply to all commands.

PR #96, opened 2017-12-28
usage: separate commands and options

darkowlzz:
This change separates command and option flags in the usage by replacing
`flag.PrintDefaults()` with `text/tabwriter` and flag lookups iterating
over command and option flags.

Output:
```
$ mage -h
mage [options] [target]

Commands:
  -init     create a starting template if no mage files exist
  -version  show version info for the mage binary

Options:
  -f     force recreation of compiled magefile
  -h     show this help
  -keep  keep intermediate mage files around after running
  -l     list mage targets in this directory
  -t     timeout in duration parsable format (e.g. 5m30s)
  -v     show verbose output when running mage targets
```

NOTE: I tried implementing this with 2 `FlagSet` but that caused some weird behavior and complicated the flow due to double parsing of arguments and overide of `-h` in one flagset, which is an default flag. This approach worked out easily. Hope this is good.

Related to #94


natefinch:
Thanks for taking this on!  I think -l should be up with the commands, though.  -h is kind of a special case... by itself it's a command, but it's also an option to show help for a target.  Maybe we could show -h in both sections. 

In this case, it's probably better to just make the usage output a hard coded string, then we can format it however we like (I often do this for small commands that I want to customize the output for).

darkowlzz:
I've moved `-h` into commands and added a manual entry in options for `-h` with proper custom description.

```
mage [options] [target]

Commands:
  -init     create a starting template if no mage files exist
  -l        list mage targets in this directory
  -h        show this help
  -version  show version info for the mage binary

Options:
  -h     show description of a target
  -f     force recreation of compiled magefile
  -keep  keep intermediate mage files around after running
  -t     timeout in duration parsable format (e.g. 5m30s)
  -v     show verbose output when running mage targets
```

vassudanagunta:
Commands should look like commands, not flags or options. I recommend changing the title of this pull request to "Distinguish commands from options", and make corresponding changes to the commits so that the result is as follows:

```
mage [command]
mage [options] [target]

Commands:
  help     show this help
  init     create a starting template if no mage files exist
  list     list mage targets in this directory
  version  show version info for the mage binary

Options:
  -h   show description of a target
  -f   force recreation of compiled magefile
  -k   keep intermediate mage files around after running
  -t   timeout in duration parsable format (e.g. 5m30s)
  -v   show verbose output when running mage targets
```



Note that I changed `-keep` to `-k` for consistency. If the full word is preferred, it should be `--keep` as is pretty standard for CLI that support both letter and word flags. But then why only `--keep` and not also `--force`, etc?  I'd say stick to the single letter, and as a future enhancement add full word alternates as synonyms.

vassudanagunta:
I just realized a possible problem with my proposal: commands and targets would be in the same namespace, meaning commands would be reserved words that can't be used as target names.  I think it is important to visually distinguish commands from flags, so some solution should be devised.

Options that I can think of:

- commands have a different prefix, e.g. `mage :version`. I don't like this. 
- running targets requires a command, e.g. `mage do -f clean build`
- commands are reserved words. The main problem with this is if a new command is added, it will force the renaming of targets that already use that name. 

Make doesn't distinguish commands from targets. I don't use make much. Does make support `make --version [targets]`?

interlock:
I would like to propose a variation on `mage :version` which is `mage mg:version`

Basically, command namespaces.

natefinch:
Here's my thought .... commands should be full words, e.g. `mage -version` and flags should be single letters, e.g. `mage -k foobar` 

I don't think there will be so many commands that we need to namespace them, and I'd rather avoid that complication to the code.  I think just name vs. single character is good enough.

Thoughts?

vassudanagunta:
I guess that works. But my idea world is that

vassudanagunta:
It works. I only wish yearn for some consistency across CLIs. Maybe that's just because I'm old. Just my 2¢. Feel free to ignore the pennies. :)

```bash
git command -o --option args
bash -o --option args
diff -o --option args
grep -o --option args
rsync -o --option args
brew command -o --option args
hugo command
hugo -o --option args
go command -o -option args    (single dash for full word flags)
```

But:
```bash
mage -o args    :)
mage -command     :(
```

All of the above, including Mage, special case `--version`, immediately executing and ignoring any other args. :)

Issue #19, opened 2017-09-29
add --clean to remove old generated binaries from the cache


natefinch:

Zanadar:
@natefinch If this isn't claimed, I'd like to take it on.
Questions:
1. Would this be a flag intended to be run on its own, or could it be included with any other command? If the latter, would it run before or after the command in question? If the former, is this a command similar to `showVersion` or `mageInit` that would get stuck in this if chain: https://github.com/magefile/mage/blob/master/mage/main.go#L91 ?
1. How do you think this should interact with `Keep` or is that command orthogonal?

natefinch:
Awesome, thanks!  

So.... ideally I'd like to refactor the command-ish flags so they are required to be run only by themselves.  As much as it might be handy to run `mage --clean foo bar`, the combinatorics become crazy as more commands are created.

So, let's say, starting with this command, if -clean is set, we'll bail if fs.NArgs() > 0 or if fs.NFlags() > 1.  I.e. bail if the user specified any other flags or positional args.

And then we can do a separate PR to make -version and -init work the same way.

natefinch:
This is done as stated above

PR #97, opened 2017-12-28
Teach mage clean

Zanadar:
Addresses https://github.com/magefile/mage/issues/19 

Starts the refactor of `-init` and `-version` into commands.


natefinch:
Thanks for this!  Is it ready for review, or are you still working?  Let me know.

Zanadar:
It should be ready for a long look. Thanks!

mage/main.go
```diff
@@ -99,33 +111,45 @@ func ParseAndRun(dir string, stdout, stderr io.Writer, stdin io.Reader, args []s
...
 		return 0
+	case CLEAN:
+		dir := mg.CacheDir()
+		if err := removeContents(dir); err != nil {
+			log.Println("Error:", err)
+			return 1
+		}
+		log.Println(dir, "cleaned")
+		return 0
 	}
```
natefinch: can you add 
```
case None:
    return Invoke(inv)
default:
   panic(fmt.Errorf("Unknown command type: %v", cmd)
}
```

That way we will make sure to catch if we add any new commands.

Also, list should be a command (the -l flag)

Zanadar: Would you mind if I kept the updating of `list` to a command as a separate PR?

Zanadar:
@natefinch Updated per your comments.  I wonder if I might leave the updating of `list` to a command to a separate PR to keep this PR focussed?

Zanadar:
@natefinch Anything I should do here to see this get to master?

mage/main.go
```diff
@@ -401,3 +439,22 @@ func RunCompiled(inv Invocation, exePath string) int {
...
 	return sh.ExitStatus(c.Run())
 }
+
+func removeContents(dir string) error {
+	d, err := os.Open(dir)
```
natefinch: why are you opening the directory instead of ioutil.ReadDir(dir) ?

mage/main.go
```diff
@@ -401,3 +439,22 @@ func RunCompiled(inv Invocation, exePath string) int {
...
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+	names, err := d.Readdirnames(-1)
+	if err != nil {
+		return err
+	}
+	for _, name := range names {
+		err = os.RemoveAll(filepath.Join(dir, name))
```
natefinch: if you use ReadDir, then you could check if something was a file before removing it, that way if someone happens to make a subdirectory in this directory, we won't wipe it out.

Zanadar:
Cool! Made some changes per your feedback. LMK what you think. Thanks for your dutiful attention to this humble PR!

natefinch:
LGTM, I'll merge once the tests pass

Zanadar:
Beautiful!

natefinch:
Thanks for all your work! :)

Issue #79, opened 2017-10-19
The mage released binary isn't self contained


bep:
```bash
Resolving deltas: 100% (29367/29367), done.
Error: failed to check types in directory: magefile.go:14:2: could not import github.com/magefile/mage/mg (cannot find package "github.com/magefile/mage/mg" in any of:
	/go/src/github.com/gohugoio/hugo/vendor/github.com/magefile/mage/mg (vendor tree)
	/usr/local/go/src/github.com/magefile/mage/mg (from $GOROOT)
	/go/src/github.com/magefile/mage/mg (from $GOPATH))
Exited with code 1
```

And while I do understand the error above and know the workaround for it, I did expect this to just work without any "go getting". I guess the released binary is just for bootstrapping and not really something one can use for added stability?

Having it in my vendor file would not work either (chicken and egg).

natefinch:
I see what you mean about chicken and egg with downloading deps used by your magefile itself. 

There's a couple possibilities here. One is to have mage dump its own libraries into the vendor directory when it runs.  Then you'd at least be guaranteed to have those and the stdlib.  That doesn't help if you want to use other libraries in your magefile, though.

Another option is that we could have mage integrate with dep to run `dep ensure` before it compiles the magefiles, then it would definitely have everything it needs.  That's a lot more heavy weight, but a lot more complete solution.

A final option would be to let you define a bootstrap magefile that gets run first, that you could guarantee only uses the stdlib, to do whatever needs to be done to set up so the rest of your magefiles have the dependencies they need.  That is most flexible, but puts the onus on the user to write the bootstrap, and adds a lot of complexity.

If you have suggestions on other ways to handle this, I'm all ears.

natefinch:
The other option is to check in the dependencies for your magefiles, so they don't need anything after git clone.

bep:
* I think in most cases you would be fine with Mage + Go stdlib. 
* Adding `dep` as a dependency isn't really solving this ... no dependency wanted issue.
* Dumping build-related stuff into the apps /vendor folder does not sound right either. These libraries have nothing to do with the app itself.
.

natefinch:
[...]

PR #98, opened 2017-12-30
sh: configure env var expansion options

### LINKED ISSUE #56: configure sh options
Auto-expand environment variables in sh should be configurable.. on or off for everything, or  include/exclude command names.  Include/exclude would be a list of command names to either turn expanding environment variables on or off for.  For example, "don't expand environment variables when I run the 'foo' command.".  Thus if you have a command that expects an argument with a $ in it, you can pass that to it without it getting auto-expanded.  

All of the above also for the upcoming auto-glob in #55

API:  
```go
func CfgExpandEnv(m CfgMode, cmds ...string) {}
func CfgGlobFIles(m CfgMode, cmds ....string) {}

var CfgMode int
const (
    OnForAll CfgMode = iota // enable the config for all commands (default)
    OffForAll               // disable the config for all commands
    IncludeCmds             // disable the config except for the given commands
    ExcludeCmds             // enable the config except for the given commands
)
```

If you pass a list of commands for OnForAll or OffForAll, we panic.  If you don't pass a list of commands for IncludeCmds or ExcludeCmds, we panics.

This would generally be used in an init function so that it takes effect for all targets... but nothing forces you to do it that way.

darkowlzz:
This change adds Configuration Mode options for expanding env vars in
commands.

fixes #56

### DIFF
--- a/sh/cmd.go
+++ b/sh/cmd.go
@@ -2,6 +2,7 @@ package sh
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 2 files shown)


natefinch:
This looks awesome, let me take a better look at it in the next day or so.

darkowlzz:
Please take your time :)

I wanted to discuss about using a context object that could be passed around from `mage` to `sh` and `mg`, or maybe have `sh` and `mg` objects, which can store some context information about any specific execution. I felt the need for this when I had to create `expandEnvConfig` and `expandEnvDefault` as global variables in `sh/cmd.go` and due to the global nature, in the tests, their values are not cleared across the tests.

I think if we create `sh` and `mg` objects in `mage`, we can solve the issue in #79 . It would make mage self contained.

I would love to hear about any other ways to do this correctly.

natefinch:
Hi, sorry this sat so long.  So, on reflection, I think this is more complicated than I'd really prefer.  Like you said, it ends up being global variables and all that.  I think that if you don't want auto-expansion, it might be best just to not use the helpers here and do it the old fashioned way.

PR #99, opened 2018-01-22
Add explicit return statements

philoserf:
The single file `build/.build.go` had a few bare returns.

This pull request will make all returns explicit. This file will be internally consistent and also consistent with the other go files. The other files already use explicit returns.


natefinch:
Thank you!   Bare returns are the devil.

PR #100, opened 2018-01-23
Improve target

philoserf:
This pull request demonstrates issues with the target package API. While I find the API desireable for use in my magefiles some improvement is required.

1. target confuses _target_, _dst_, and _sources_. It places them in different orders in different places or gives them different names when they are the same thing.
2. the common case when the target does not exist is not yet provided for.

Three commits are provided. The diff logic on the PR makes this look like changes. There is no change to `target.go` or to the existing test. The existing test is renamed to indicate its use as a test of target package internals.


magefile.go
```diff
@@ -14,6 +14,11 @@ import (
 	"github.com/magefile/mage/sh"
 )
 
+// Runs all the golang tests while checking for race conditions
```
philoserf: Offered as a useful task perhaps should be packaged in a real pull request adding common golang project mage tasks.

target/target_internal_test.go
```diff
@@ -0,0 +1,63 @@
+package target
```
philoserf: This is the original `target_test.go` renamed and unchanged.

target/target_test.go
```diff
@@ -1,63 +1,37 @@
-package target
+package target_test
```
philoserf: test target from the outside thus using it as a magefile would.

target/target_test.go
```diff
@@ -1,63 +1,37 @@
...
-	}()
-
-	table := []struct {
-		desc    string
-		src     string
-		targets []string
-		expect  bool
+	tests := []struct {
+		name    string
+		dst     string
```
philoserf: The _target_ of the `target package` is called _dst_ by the API. This may be clearer if the first argument where _tgt_.

target/target_test.go
```diff
@@ -1,63 +1,37 @@
...
-		v, err := Path(c.src, c.targets...)
-		if err != nil {
-			t.Fatal(err)
-		}
-		if v != c.expect {
-			t.Errorf("%s : expecting %v got %v", c.desc, c.expect, v)
-		}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := target.Path(tt.dst, tt.sources...)
```
philoserf: This call to `target.Path` will always fail in the std library call to `stat` as currently written in the target package when the target does not exist.

target/target_internal_test.go
```diff
@@ -0,0 +1,63 @@
...
+		{"Files only target doesn't check", "testdata/target_file",
+			[]string{"testdata/src_file"}, false},
+		{"Directories only target checks", "testdata/src_dir",
+			[]string{"testdata/target_dir"}, true},
+		{"Directories only target doesn't  checks", "testdata/target_dir",
+			[]string{"testdata/src_dir"}, false},
+	}
+
+	for _, c := range table {
+		v, err := Path(c.src, c.targets...)
```
philoserf: notice that the original test reverse _src_ and _targets_

target/target_test.go
```diff
@@ -1,63 +1,37 @@
...
+			if err != nil {
+				t.Errorf("Path() error = %v", err)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("Path() = %v, want %v", got, tt.want)
+			}
+		})
 	}
 }
```
philoserf: DO NOT MERGE

philoserf:
here is the test failure I have introduced to illustrate the missing target issue:

`    	target_test.go:29: Path() error = stat dst: no such file or directory`

I will dig in to see about adding the missing target capability that will make this pass while keeping all the internal tests passing.

philoserf:
It'll be here to review. No need to keep it open as It was not intended for merge.

natefinch:
Thanks, I'll take a look.  It definitely is a problem.

PR #104, opened 2018-02-10
Add alternative way of adding alias.

gernest:
Any word following  alias: in a comment on a mage target is  an alias.

example

```go
//+build mage

package main

import (
	"fmt"
)

// prints hello,word on stdout
// alias: hello
func helloWorld() {
	return fmt.Println("hello, world")
}
```

This will register hello as an alias to helloWorld target.

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -1,6 +1,7 @@
 package parse
 
 import (
+	"bufio"
 	"fmt"
 	"go/ast"
 	"go/build"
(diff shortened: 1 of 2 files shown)


gernest:
I have been using this for sometime on my local machine. I think others might find this useful , it has helped me alot keeping my magefiles sane.

gernest:
Oops! Was trying to comment but accidentally closed

natefinch:
I'm wary of adding our own customized tags in comments.  I like that it keeps the alias close to the definition of the function, but it seems like it could be easy to screw up, and might collide with other similar tags.  Also it complicates the processing of the documentation.

I'd like to see if there's another way to do this that isn't in the comments....

gernest:
I see, I have been using this everyday can't think of another way more convenient to me.

I will close this for now.

PR #115, opened 2018-04-04
Add `-compile` flag to compile a static binary

jen20:
This commit adds a new flag, `-compile`, which takes the argument of the path to which to compile a static binary, but not execute it. This is useful if you wish to run tasks on a machine which does not have Go available.

Examples:

```shell
$ mage -compile ./test

$ file ./test
./test: Mach-O 64-bit executable x86_64

$ nm ./test | grep Install
00000000010a8b50 t main.Install
```

```shell
$ GOOS=freebsd mage compile ./test

$ file ./test
test: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), statically linked, with debug_info, not stripped

$ nm ./test | grep Install
00000000004a7570 T main.Install
```

### DIFF
--- a/mage/command_string.go
+++ b/mage/command_string.go
@@ -2,15 +2,15 @@
 
 package mage
 
-import "fmt"
+import "strconv"
 
-const _Command_name = "NoneVersionInitClean"
+const _Command_name = "NoneVersionInitCleanCompileStatic"
 
-var _Command_index = [...]uint8{0, 4, 11, 15, 20}
+var _Command_index = [...]uint8{0, 4, 11, 15, 20, 33}
 
 func (i Command) String() string {
 	if i < 0 || i >= Command(len(_Command_index)-1) {
-		return fmt.Sprintf("Command(%d)", i)
+		return "Command(" + strconv.FormatInt(int64(i), 10) + ")"
 	}
 	return _Command_name[_Command_index[i]:_Command_index[i+1]]
 }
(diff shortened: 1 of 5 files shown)


mage/command_string.go
```diff
@@ -2,15 +2,15 @@
 
 package mage
 
-import "fmt"
+import "strconv"
```
natefinch: interesting, we must be using different versions of stringer.

site/content/index.md
```diff
@@ -53,18 +53,22 @@ plugin.  Every tool you use with Go can be used with Magefiles.
 ```
 mage [options] [target]
 Options:
-  -f    force recreation of compiled magefile
-  -h    show this help
+  -clean
```
natefinch: thanks for fixing this up... I should write a test to make sure it stays up to date.

mage/command_string.go
jen20: Ah, I was using `71657689` for `golang.org/x/tools/cmd/stringer`, I just updated to the latest `master`, `ac136b6c`, and see the same code generated, so I assume this is what `stringer` is generating right now.

site/content/index.md
jen20: Probably best for another thread rather than directly related to this change, but it's likely possible to generate the documentation examples from the actual running program?

natefinch:
There's one sorta-bug that I found, but it can be worked out later. And that's the fact that the mage binary checks for the MAGE_VERBOSE flag and passes that in to the running binary... and the running binary doesn't do that itself.  But that's fixable, and not critical.

Issue #119, opened 2018-06-22
SubCommands


natefinch:
Sometimes you want to have subcommands, like `mage build:foobar` and `mage build:baz` where foobar and baz are *subcommands* of the build command.

A relatively straightforward way to declare this could be thusly:
```go
// somewhere in the magefile/mage/mg package
type Command struct{}

// in your magefile

type Build mg.Command

func (Build) Foobar() error {
    // do your foobar build
    return nil
}

func (Build) Baz() error {
    // do your baz build
    return nil
}
```

natefinch:
fixed by #121

PR #121, opened 2018-07-13
Add namespace/subcommand support to Mage

ryanfaerman:
This delivers #119 as it is something that I wanted to see available. I've never worked with the `go/ast` package before, so I may be doing some things in a less than ideal way.

This enables the nearly the exact syntax outlined in #119 to work.

```go
type Build mg.Namespace

func (Build) Foobar() error {
    // do your foobar build
    return nil
}

func (Build) Baz() error {
    // do your baz build
    return nil
}
```
With that, the functions are now exposed in a namespaced fashion:

```
$ mage -l
Targets:
  build:foobar
  build:baz
```

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -28,22 +28,36 @@ type PkgInfo struct {
 // Function represented a job function from a mage file
 type Function struct {
 	Name      string
+	Receiver  string
 	IsError   bool
 	IsContext bool
 	Synopsis  string
 	Comment   string
 }
 
+// TemplateName returns the invocation name, supporting namespaced functions.
+func (f Function) TemplateName() string {
+	if len(f.Receiver) > 0 {
+		return strings.ToLower(fmt.Sprintf("%s:%s", f.Receiver, f.Name))
+	}
+
+	return f.Name
+}
+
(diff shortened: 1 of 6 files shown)


parse/parse.go
```diff
@@ -28,45 +28,59 @@ type PkgInfo struct {
...
+
+	return f.Name
+}
+
 // TemplateString returns code for the template switch to run the target.
 // It wraps each target call to match the func(context.Context) error that
 // runTarget requires.
 func (f Function) TemplateString() string {
+	name := f.Name
+	if len(f.Receiver) > 0 {
```
xxxtonixxx: Here too :+1: https://github.com/magefile/mage/pull/121#discussion_r202502819

ryanfaerman:
I'd be happy to rebase the last few commits if you'd like. I just didn't want to do that without checking first.

natefinch:
@ryanfaerman btw the website link on your github profile doesn't work (went looking for a twitter handle so when I announce this I can credit you :) )

ryanfaerman:
@natefinch Thanks for the heads up on my website not working (I've fixed the link). I'm not really on twitter or anything like that. Glad I could help improve mage!

Issue #122, opened 2018-07-16
It is difficult to get make-like behaviour with target.Path


bdalton:
There are two cases in make where a rule is executed:
1. the target does not exist.
2. the target is older than one of the sources.

target.Path(dst, sources...) returns true, nil on (2), but returns false, err on (1).
That leads me to think that maybe I could do something like this:

```
outOfDate, err := target.Path(dst, source1, source2)
if err != nil && !os.IsNotExist(err) {
   return err
}
if outOfDate || os.IsNotExist(err) {
   ... action ...
}
```

Unfortunately, if any of the sources doesn't exist, I will also get an os.IsNotExist() also returning true so I can't distinguish between the target not existing and one of the sources existing.

bdalton:
Sure, I'll submit a PR later today

bdalton:
PR submitted

natefinch:
Thank you!

natefinch:
fixed by 131.  If there's a different behavior you'd like, we can add it to another function, perhaps.

PR #123, opened 2018-07-18
Fix for Issue #122

bdalton:
Love magefile and will be using it for all my go projects going forward.

This started as my own target module when I couldn't get the bundled 'target' module to work the way I needed it to. I've expanded the unit tests also. Feel free to cherry-pick or alter anything.

Barnaby

### DIFF
--- a/target/target.go
+++ b/target/target.go
@@ -7,11 +7,14 @@ import (
 	"time"
 )
 
+// Path is a synonym for File
 var Path = File
 
-// Path reports if any of the sources have been modified more recently
-// than the destination.  Path does not descend into directories, it literally
-// just checks the modtime of each thing you pass to it.
+
+// File returns true if either the target does not exist or
+// is older than at least one of the sources. If any of the
+// sources are a directory, the modification time of the newest
+// file within the directory or any of its children is used.
 func File(target string, sources ...string) (bool, error) {
 	sourcesModTime, err := checkDependencies(sources)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


natefinch:
I've been looking at this off and on.  It's a bit bigger than I hoped, which is why I haven't gotten to it before now.  I'll try to get a handle on it soon.

natefinch:
Oh man, I just started looking at the target stuff and whoo boy, it's broken.  Lemme look at your code.

natefinch:
Clarification: the existing code is broken.... not yours :)

natefinch:
#131 addresses this.  If there's something different you'd like to do, we can talk about that (it looks like you changed the behavior slightly)... but I think that would need to perhaps be a new function.

bdalton:
The tests codified what I needed. I couldn't get that behavior even with extra checks with the old code.
If it passes those, then it works for me. I'm not attached to the code I wrote.

I'm in the middle of a release now, so am hesitant to upgrade something I don't need to. I'll update to your changes and stop using my patch after I get things out the door.

Thanks

natefinch:
Cool, yeah, like I said, I added your tests temporarily to my branch, and
they passed.  I should have just left them in there, honestly (more tests
is better), but alas, hindsight.

Anyway, yes, I understand about releases, no worries.  I merged my PR in,
because I was pretty confident in it.   Thanks again for the contribution,
it definitely spurred me to look more at this, and I'm glad I did.

PR #128, opened 2018-07-25
Log dependencies in runDeps

### LINKED ISSUE #110: -v should print out dependencies that are run
if you have 
```go
func Foo() {
    mg.Deps(Bar, Baz)
}
```
and you run `mage -v foo` currently we print out 
```
Running Foo
```

But it would be nice if it said
```
Running Bar
Running Baz
Running Foo
```

dotpy3:
Closes #110.
With this commit, `mage` prints executed dependencies when using `-v`:

```
$ mage -v buildWithDeps
Running target: BuildWithDeps
Running dependency: Build
```

### DIFF
--- a/mg/deps.go
+++ b/mg/deps.go
@@ -3,6 +3,8 @@ package mg
 import (
 	"context"
 	"fmt"
+	"log"
+	"os"
 	"reflect"
 	"runtime"
 	"strings"
@@ -75,6 +77,7 @@ func runDeps(ctx context.Context, fns ...interface{}) {
 	for _, f := range fns {
 		fn := addDep(ctx, f)
 		wg.Add(1)
+		logger := log.New(os.Stderr, "", 0)
 		go func() {
 			defer func() {
 				if v := recover(); v != nil {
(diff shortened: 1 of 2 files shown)


dotpy3:
Any news about this? :)

mg/deps.go
```diff
@@ -89,6 +92,9 @@ func runDeps(ctx context.Context, fns ...interface{}) {
 				}
 				wg.Done()
 			}()
+			if os.Getenv("MAGEFILE_VERBOSE") != "" {
```
natefinch: this is defined in the Verbose() method, so we can reuse that -  `if Verbose() {`

mg/deps.go
```diff
@@ -151,10 +159,17 @@ func name(i interface{}) string {
 	return runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
 }
 
+func displayName(name string) string {
```
natefinch: I think it might be useful to print the package, actually.  Targets get printed without the package name, but by definition they have to be in the main package. Dependencies don't.    Maybe if the package is "main" then don't print the package, but otherwise, do?

dotpy3: Good remark, I was mostly thinking of `main` indeed!

dotpy3:
I've added a test for it, and addressed the comments!

PR #130, opened 2018-08-13
Add sh.Copy

lpar:
It's only a 10 line helper function, but it's 10 lines that I suspect people often want and often get subtly wrong.


natefinch:
Three things... 

First, can you switch the arguments?  While I personally agree with left to right, most go functions do the opposite (for example, `io.Copy(dst io.Writer, src io.Reader)`) .  The rationale is that it's like variable assignment i.e. `dst := src`.  But for our purposes, just being consistent with io.Copy is reason enough.  (also can you make them dst and src?  That's also a standard.)

Second, I think you'll want to copy file mode from the source as well, so open the destination with 
`to, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, from.Mode())`

lpar:
Good point re: reversed args. I think the desire to preserve permissions makes it a more compelling addition too.

lpar:
OK, I've added tests for Copy, and also for Rm which didn't look like it had any. (140 lines of test code for 39 lines of helper code, robust testing is hard!)

natefinch:
You, sir, are a saint :)

sh/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}()
+	srcname, err := makeTestFile(mytmpdir, "test1.txt")
+	if err != nil {
+		t.Error(err)
+	}
+	destname := filepath.Join(mytmpdir, "test2.txt")
+
+	t.Run("sh/copy", func(t *testing.T) {
+		cerr := sh.Copy(destname, srcname)
```
natefinch: so, I don't think it's super useful to name errors something other than `err` unless it's actually necessary to distinguish between them (and it's not here).  It kind of makes the code harder to read, because when I see `if cerr != nil {` I have to think about what type `cerr` is.  Whereas `if err != nil {` might as well be one complete token, because it's so common in go, everyone knows what it means immediately.

sh/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/magefile/mage/sh"
+)
+
+func makeTestFile(dir string, name string) (string, error) {
```
natefinch: so you can probably replace this with [ioutil.WriteFile](https://golang.org/pkg/io/ioutil/#WriteFile).  I don't think the random contents are critical to the test.

sh/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	f1bytes, err := ioutil.ReadFile(file1)
+	if err != nil {
+		return fmt.Errorf("can't read %s: %v", file1, err)
+	}
+	f2bytes, err := ioutil.ReadFile(file2)
+	if err != nil {
+		return fmt.Errorf("can't read %s: %v", file2, err)
+	}
+	cmp := bytes.Compare(f1bytes, f2bytes)
```
natefinch: use bytes.Equal instead, so you don't have to compare vs 0

sh/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+		return fmt.Errorf("files %s and %s have different contents", file1, file2)
+	}
+	return nil
+}
+
+func TestHelpers(t *testing.T) {
+
+	mytmpdir, err := ioutil.TempDir(os.TempDir(), "mage")
+	if err != nil {
+		t.Errorf("can't create test directory: %v", err)
```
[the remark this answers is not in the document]
lpar: Aha, I thought Fatalf would be fatal, i.e. stop all testing.

sh/helpers_test.go
lpar: I do it to avoid warnings about variable shadowing when I lint the code.

sh/helpers_test.go
lpar: I think I disagree with that. I can imagine situations where copy was implemented using a fixed size buffer and a fencepost error caused corruption for specific file sizes.

natefinch: I think that if we want to do a bunch of tests with different contents, that's fine, but doing one test with random contents means that we won't know what caused the problem if the test fails.  Also, we're using io.Copy, so I'm not too worried about the actual copying of contents being wrong :)

sh/helpers_test.go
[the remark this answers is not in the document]
natefinch: Yeah, it took me forever to internalize the difference between error and fatal...

sh/helpers_test.go
natefinch: IMO variable shadowing isn't a bad thing.  Yes it can sometimes cause errors, but that's pretty rare in my experience, and it makes the code a lot easier to read precisely because you don't need to give every error a different name.  Then I don't have to think about if this is cerr or derr... it's just always err.  ¯\\_(ツ)_/¯ .  Your mileage may vary.

sh/helpers_test.go
lpar: OK, if you're happy trusting io.Copy fair enough.

natefinch: if io.Copy was broken, someone would have figured it out long before us :)

natefinch:
Thank you!

PR #131, opened 2018-08-23
fix target library

natefinch:
There were a few different problems with this library, which were a lot clearer
with added tests. The main change is to fix it such that if the target is missing
we automatically call it as needing to be regenerated.  Plus target.Path was 
just plain broken.


natefinch:
fixes #122

PR #137, opened 2018-08-26
Check for duplicates with namespace prefixed if any

### LINKED ISSUE #136: Build target uniqueness constraint
Currently all build targets must be unique regardless of namespace. This kind of limits the use of namespaces.

For example something like below is not possible:

`init:instance`
`build:instance`

I actually wanted to make this work such that following is possible. This requires #135 

`build:instance inst1`

faxal:
Fixes #136


natefinch:
Ahh, I see the problem: https://github.com/magefile/mage/blob/master/parse/parse_test.go#L9

The test manually selects which files to use, rather than using mage to figure it out via the build tags, so the fact that the build tag was messed up didn't matter.

Issue #140, opened 2018-08-28
Mage fails to locate transient dependencies saved via Go modules


dradtke:
It looks like Mage doesn't fully support dependencies defined by Go modules. In particular, transient dependencies fail because they're not placed in the place that Mage expects.

To reproduce, create a new project containing a dependency, followed by a Magefile that uses it:

```
$ go mod init app
$ go get golang.org/x/text/unicode/norm
```

```go
// +build mage

package main

import (
        "golang.org/x/text/unicode/norm"
)

func Test() {
        print("unicode version: " + norm.Version)
}
```

Running `mage test` fails with the following error:

```
Error: failed to check types in directory: magefile.go:6:2: could not import golang.org/x/text/unicode/norm (type-checking package "golang.org/x/text/unicode/norm" failed (/home/damien/go/pkg/mod/golang.org/x/text@v0.3.0/unicode/norm/normalize.go:15:2: could not import golang.org/x/text/transform (cannot find package "golang.org/x/text/transform" in any of:
        /usr/local/go/src/golang.org/x/text/transform (from $GOROOT)
        /home/damien/go/src/golang.org/x/text/transform (from $GOPATH))))
```

This is because `golang.org/x/text/unicode/norm` depends on `golang.org/x/text/transform`, which exists, but not in the GOPATH. Using modules, Go downloads both dependencies to `~/go/pkg/mod/golang.org/x/text@v0.3.0/`.

Mage should be updated to include the logic utilized by e.g. `go list -m -f '{{.Dir}}' all` to locate modularized dependencies, if being compiled with Go 1.11 or later.

natefinch:
I'm looking at this now, thanks for letting me know.

dradtke:
I just noticed https://github.com/magefile/mage/issues/124, so maybe Mage works with 1.11 in general and only fails with these transitive dependencies.

natefinch:
Russ fixed the fact that they had broken people who relied on parsing source but it sounds like there's a more subtle bug that is likely my fault.  I'll look into it.

natefinch:
Oh... make sure you use mage built with go 1.11.  te 2.2.0 release is that, or of course you can build it yourself.  Note that if you use mage's `mage install` target, it will build itself with your default go installation.

dradtke:
I can confirm that this is still an issue with the 2.2.0 release, as well as Mage built from source using Go 1.11.

natefinch:
[...]

PR #144, opened 2018-09-01
Use go/build from 1.11 to support modules

natefinch:
This updates the magefile/build code to use go/build from 1.11, which has some needed
tweaks to fully support modules


natefinch:
fixes #140

PR #153, opened 2018-09-09
massive cleanup using go list

natefinch:
This removes the custom build folders and the copy of source importer.  We
use go list now instead of the build pacakge, and just diff go list with and
without the mage tag to find the files with a mage tag.  We also don't need
as much type info for building, so we don't need source importer.


johnrichardrinehart:
Nice job!

PR #160, opened 2018-09-12
Add support for bash completion

### LINKED ISSUE #113: Tab completion
Mage is so amazing, a breath of fresh air for build tools! Love the minimalism and potential for highly cross-platform builds! Hey, could we get some tab completion for target names? Perhaps a bash completion script?

frou:
Hi there - give this a try. It can be loaded by adding this to one's `.bash_profile`:

`. "$GOPATH/src/github.com/magefile/mage/bash_completion"`

---

Mage's own magefile could probably have a target to make the above step unnecessary. It would have to find the system's `bash_completion.d` directory (e.g. `/usr/local/etc/bash_completion.d`) and create a symlink in it to our script, which would result in automatic loading.

Fixes #113

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1,7 +1,7 @@
 _mage() {
 	# Get the partial word typed by the user.
 	local PARTIAL_WORD
-	PARTIAL_WORD="$(echo "$2" | tr '[:upper:]' '[:lower:]')"
+	PARTIAL_WORD="$2"
 
 	# Narrow in on a category of completions to offer.
 	local CATEGORY
--- a/bash_completion
+++ b/bash_completion
@@ -1,0 +1,28 @@
+_mage() {
+	# Get the partial word typed by the user.
+	local PARTIAL_WORD
+	PARTIAL_WORD="$(echo "$2" | tr '[:upper:]' '[:lower:]')"
+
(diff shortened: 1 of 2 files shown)


natefinch:
Works well on my machine.  I kinda feel like this might belong better as docs on the website... I worry that trying to deploy this automatically is going to be more complexity than I really want in Mage itself.

frou:
Perhaps fully automatic deployment of completion wouldn't make sense unless mage was being installed using an OS-level package manager (rather than `go get`).

How about the this, and the user can just run it if they want.

(I think if the process were even more manual than this, the number of people who would actually end up having completion would absolutely plummet).

natefinch:
Yeah, the problem is that we'd have to do a lot of work to know all the edge cases of if that directory can live somewhere else.... what happens if they use a different shell...  

I'd rather have this as a good example that most people can just drop in place, and not have to try to support it when it fails for some weird OS/shell/whatever. You know there's going to be some linux person who has their bash completion directory set up on like a network share that requires a password to log into it or something crazy like that.

I like bash completion, but I don't feel like it's critical for mage.   You can easily make aliases in your magefile to make commands as short as you like....

PR #162, opened 2018-09-14
consider missing cache dir cleaned

graphaelli:


natefinch:
Smart, thanks.  Can you add a quick test for this?

graphaelli:
I thought I might get away without one for one this simple, looking for a way to set the cache dir for a test now...

natefinch:
That test isn't doing what you think it's doing.   Instead of readDir you need to remove the cache directory (it's ok, mage will recreate it as needed).

Something like this:

```
	if err := os.RemoveAll(mg.CacheDir()); err != nil {
		t.Fatal(err)
	}
	code := ParseAndRun(ioutil.Discard, ioutil.Discard, &bytes.Buffer{}, []string{"-clean"})
	if code != 0 {
		t.Errorf("expected 0, but got %v", code)
	}
```

I tested it and this fails before your change and passes after. :)

graphaelli:
I was looking for a way to set the cache dir just for the test and was going to pass `path.Join(mg.CacheDir(), "foo")` into that but didn't actually find a way as written, didn't mean to push this up.  Since you're good with that `RemoveAll` I'll just put that in.

PR #172, opened 2018-09-19
Addresses file move error in Docker.

bfitzsimmons:
Addressed the error I was experiencing in Docker containers where copying to, or from, the cache dir. was resulting in an error.

Example:
`DEBUG: 19:26:06.869194 error caching main file:  rename /go/src/github.com/<my>/<repo>/mage_output_file.go /root/.magefile/mainfiles/516a98d0b7cb5d2401fd03b8abe3ebdb5f72ba26: invalid cross-device link`

- This also addressed a couple instances of variable shadowing (thanks `go vet`).
- Aaaaand, a trailing space is removed. :)

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -578,7 +578,7 @@ func useExistingMain(cachedMain, path, hash string) bool {
 		debug.Printf("error copying cached mainfile from %s to %s: %v", cachedMain, path, err)
 	}
 	// Remove the file.
-	if err = os.Remove(cachedMain); err == nil {
+	if err = os.Remove(cachedMain); err != nil {
 		debug.Println("error removing cached mainfile from cachedir: ", err)
 		return true
 	}
--- a/mage/main.go
+++ b/mage/main.go
@@ -187,7 +187,7 @@ Commands:
   -version  show version info for the mage binary
 
 Options:
-  -d <string> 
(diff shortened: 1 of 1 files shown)


natefinch:
You know, I wondered if that file move was going to screw me with different drives.  Here's the thing... the whole reason we cache the generated main file is so the mod time doesn't change.  The go tool's logic for determining when it needs to rebuild evidently includes the mod time of the files it is compiling.   So even if the contents are identical, it'll still rebuild if the file is newer than the binary it produced.  So that's why I was just caching away the mainfile.

It occurs to me (now) that I could just spoof the modtime.... I wonder if that would work.  Then I wouldn't have to cache the mainfile at all (which would be nice, because it's kinda gnarly).

natefinch:
So... the file copy you implemented will fix the error from running in docker, but it means we'd always rebuild, which adds like 500ms to the runtime of every mage command  😢 .   

I think the modtime might work... lemme give it a shot locally.

bfitzsimmons:
Sounds good. Thanks for taking a look at it @natefinch .

natefinch:
I think this will do it: https://github.com/magefile/mage/pull/173

natefinch:
Fixed in #173 :)

PR #173, opened 2018-09-19
Don't cache mainfiles, just set modtime

natefinch:
So the whole caching mainfiles thing is silly... if we want an old modtime, just set an old modtime.


natefinch:
Hmm looks like go 1.7 doesn't like chtime...

PR #174, opened 2018-09-20
fix error handling when failing to list magefiles

zbindenren:
This pull request fixes error handling. It changes the message from:

```
```

to:
```
Error determining list of magefiles: failed to list non-mage gofiles: exec: "go": executable file not found in $PATH
```

when for example the go binary is not found in PATH.


natefinch:
Awesome, thank you!

PR #185, opened 2018-10-01
Update installation instruction

ashishmohite:
`go install github.com/magefile/mage` works fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ cd $GOPATH/src/github.com/magefile/mage
 go run bootstrap.go
 ```
 
+or 
+
+```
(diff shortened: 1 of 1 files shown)


natefinch:
so, it works fine except that then `mage -version` doesn't print out version info for mage, which can be a useful debugging tool.   `go run bootstrap.go` is nearly as easy, but does embed the version info.

ashishmohite:
OK 
Got your point

Shall i close this PR ?

natefinch:
Sure.  Thank you for your input, though.  I definitely like getting updates to unclear or incorrect documentation :)

Issue #188, opened 2018-10-12
Magefile can't use application code defined in the same directory


dradtke:
I have an application whose design is inspired by [this post](https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1); in particular, domain objects are all stored in the application's root package, which is also the top-level folder of the repository. Since it's the top of the repository, I also have my main Magefile defined there, and lately it's been failing to compile because it finds two packages in that directory: `main` and (for example) `app`:

```
Mage.go:20:2: found packages main (Mage.go) and app (app.go) in /Users/dradtke/Workspace/app
```

This used to work, so I'm not sure what recent change actually caused it to break. The version of Mage I'm using:

```
$ mage -version
Mage Build Tool v1.7.0
Build Date: 2018-10-12T10:40:22-05:00
Commit: 0f9693c
built with: go1.11
```

The workaround seems to be pretty simple, which is to move application code down one level, leaving the Magefile as the sole Go file at the root. This is kind of annoying though, since it requires re-architecting the application in order to avoid compilation errors.

natefinch:
Can you run mage -l -debug in that directory and paste that here?  This sounds like a bug, probably introduced with the changes I made for 1.7.

dradtke:
Sure thing. Here's a simple case, with just these two files saved in `github.com/magefile/test`:

```go
// +build mage

// Mage.go
package main

import (
        "fmt"
        "github.com/magefile/test"
)

func Run() {
        fmt.Println(test.Message())
}
```

```go
// app.go
package app

func Message() string {
        return "success!"
}
```

```
DEBUG: 11:06:33.830955 getting all non-mage files in .
DEBUG: 11:06:33.856038 found non-mage files app.go
DEBUG: 11:06:33.856063 marked file as non-mage: "app.go"
DEBUG: 11:06:33.856066 getting all files plus mage files
DEBUG: 11:06:33.877148 time to scan for Magefiles: 46.197927ms
DEBUG: 11:06:33.877165 found magefiles: Mage.go
DEBUG: 11:06:33.898578 output exe is  /Users/dradtke/.magefile/b03ddfebe208872194e13c2a2576ad2e85f8c7b6
DEBUG: 11:06:33.898604 running go env GOCACHE
DEBUG: 11:06:33.920920 build cache exists, will ignore any compiled binary
DEBUG: 11:06:33.920942 parsing files
DEBUG: 11:06:33.921284 found target Run
[...]

PR #189, opened 2018-10-12
Fix mixed package imports

natefinch:
When I added mage:imports I added code to run the compile with -tags=mage... that was
a mistake. We don't actually want to do that because if you import code from a
directory which has both magefiles and non-mage files, then it'll fail to compile
if they're different package names.


natefinch:
Fixes #188

PR #199, opened 2018-11-18
Add mage flags to the compiled mage binary (#180)

### LINKED ISSUE #180: Add mage flags to the compiled mage binary
When I run mage I can do the following:

```
mage -l
Targets:
  build      Builds the holidays binary.
  deps       Installs go dependencies.
  docker     Build Docker image.
  test       Run all go tests.
```

When I compile the mage file I do not have the mage flags anymore:

```
mage -compile mg
./mg -l
Unknown target specified: -l
./mg -v deps
Unknown target specified: -v
```
Would be great if one can still use the flagse (-v for example).

denisbrodbeck:
Adding following flags:

* -v (verbose)
* -l (list)
* -h (help)
* -t (timeout)
* -ignoredefault

fixes #180

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -7,6 +7,7 @@ package main
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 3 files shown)


denisbrodbeck:
Hmm, that travis-ci build is certainly strange.
Go 1.10 + 1.11 are fine, Go 1.9 - 1.7 fail with missing permissions on /tmp/randomfilename.
How should I circumvent such behavior?

natefinch:
Thanks very much for this PR!  I'll take a look tonight. Not sure what's going on with the test... it's very weird that it works on 1.10+ and not on 1.9-  ... I wouldn't think that would change how the permissions on a temp file are created

mage/template.go
```diff
@@ -20,8 +21,57 @@ import (
...
+		// default flag set with ExitOnError and auto generated PrintDefaults should be sufficient
+		flag.BoolVar(&args.Verbose, "v", false, "show verbose output when running mage targets")
+		flag.BoolVar(&args.List, "l", false, "list mage targets in this directory")
+		flag.BoolVar(&args.Help, "h", false, "print out help for a specific target")
+		flag.BoolVar(&args.IgnoreDefault, "ignoredefault", false, "ignore the default build target")
+		flag.DurationVar(&args.Timeout, "t", 0, "timeout in duration parsable format (e.g. 5m30s)")
+		flag.Parse()
+		args.Args = flag.Args()
+		// With golangs flag package we can't distinguish, whether a flag's default value has been:
+		//  * provided by the user or
```
natefinch: So, the way to do the defaults to environment variables is that when you declare the default up above, instead of hard coding the default to false, get the environment variable's value, and use that as the default.

natefinch:
I am working on some updates to this PR, will push them when they're done, then hopefully the build error will go away.

natefinch:
Hope you don't mind, I tweaked it just a bit... removed the ignoredefault flag (it's basically the same as -l).  I also made a custom help usage output, because I'm picky.  Thanks super much for getting this done, and twice as much for the tests, that's just amazing.

natefinch:
merged in #201. Thanks!

natefinch:
BTW, seems like the build problem was that you can't execute files from the temp directory on more closely secured linux environments, which Travis uses.  I'm not exactly sure how Go 1.10 changes the behavior, but I saw some issues about it in the go repo.

PR #200, opened 2018-11-21
revise mage build instructions for go 1.11

swdunlop:
As of Go 1.11, the Mage bootstrap instructions are broken, because `go get -u -d` will leave the source in a read-only `$GOPATH/mod` directory instead of `$GOPATH/src`.  It should be conceptually simpler to clone the repo by hand and go run the bootstrap from there.

Well, simpler, until Go changes how things are built from online sources again..

### DIFF
--- a/site/content/index.md
+++ b/site/content/index.md
@@ -16,12 +16,12 @@ and above (possibly even lower versions, but they're not regularly tested).
 Install mage by running 
 
 ```plain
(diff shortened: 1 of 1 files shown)


natefinch:
Thanks!  

Can you make it into two sections, one for modules and one for non-modules.  because go get still works correctly if you're in your gopath

PR #207, opened 2018-12-12
Fix documentation for `GoCmdEnv`

ihgann:
`GoCmdEnv` documentation was a copy of the verbose mode documentation - this change specifies the intention of the `-gocmd` flag.

### DIFF
--- a/mg/runtime.go
+++ b/mg/runtime.go
@@ -19,8 +19,8 @@ const VerboseEnv = "MAGEFILE_VERBOSE"
 // debug mode when running mage.
 const DebugEnv = "MAGEFILE_DEBUG"
 
-// GoCmdEnv is the environment variable that indicates the user requested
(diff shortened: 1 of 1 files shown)


ihgann:
I had a further question that might be better tracked as an issue - I use my magefile with some specific targets that utilize the `go` runtime, and in particular it's a bit confusing that setting `-gocmd` does not actually persist when I invoke `mg.GoCmd()` within my own magefile.

That is to say, `-gocmd` specifically only applies to *compiling* the magefile, but this seems misleading. I think if `-gocmd` is applied, then the magefile ought to be run with var `MAGEFILE_GOCMD` set as well. That doesn't apply to this change per se, but I only noticed this documentation note because I was confused by this behaviour.

natefinch:
That's a good point about it not getting passed through. Definitely worth an issue and something that is easy to fix.

Issue #24, opened 2017-10-01
auto-args for targets?


natefinch:
It wouldn't be too hard to auto-generate a CLI for each target so the user could pass in positional args or flags per arg.

e.g.

```go
func Build(OS string, ARCH string) error {
     // build with GOOS and GOARCH
}

$ mage build -os darwin -arch amd64
building with GOOS=darwin GOARCH=amd64
```

natefinch:
May conflict with #34

jrnt30:
While it's a little ugly, I think you could support both features if you kept the multi-target support to be explicitly delimited.  The available bind targets then for the method arguments would be the union of the two.  If both targets had an `OS string` in them, they would both be passed the same value.

natefinch:
I see . what you mean, like if build had an `OS string` argument, and compile did too, but compile also had a `output string` then you'd call it like

mage build compile -OS=windows -output=foo.exe 

That's not terrible :)

vassudanagunta:
> mage build compile -OS=windows -output=foo.exe

> While it's a little ugly

That's not ugly, it's beautiful! And elegant if targets declare a same named arg and you only have to specify it once on the command line. Or where you thinking of some other approach, @jrnt30?

jrnt30:
Now that was the general thought I had a well

sean-:
I don't know about coupling the variables to the positional args for a given target.  Wouldn't most use cases be satisfied with something closer to:

```go
type Args map[string][]string

func Build(args Args) error {
  switch {
  case args.Get("GOOS") && args.Get("GOARCH"):
    // build with GOOS and GOARCH
  default:
    // other
  }
}

// Get returns the first argument
func (a Args) Get(key string) string { ... }
```

The interface is basically ripped off from https://golang.org/pkg/net/url/#Values.Get  ?  This solves a huge list of problems on our end with mage and would be a very, very welcome feature.  Maybe return the args as part of a `context.Context` so we don't need to change the function signature for anything.  ?

sean-:
That's true, too.  I think if you have multiple targets, you have to scan from left to right and parse the args between the targets and use that as the rough function signature.  Does order matter?  That's problematic UX.  How do you pre-register an argument so it's available as help?  What if a target has 10 flags available and half of them are optional, then what?  Pass them all in as arguments to the target `func`?  Use a pointer to a string?

```go
func Foo(barRequired string, bazOptional *string) error {
```

[...]

PR #210, opened 2018-12-13
Pass MAGEFILE_GOCMD to compiled binary

### LINKED ISSUE #208: -gocmd should pass MAGEFILE_GOCMD to the Magefile
As mentioned in https://github.com/magefile/mage/pull/207, if a user sets `-gocmd` when running their magefile, the go runtime they pass in is not actually available to the Magefile. It seems, instead, the setting is only used for compiling the magefile, but is not actually transferred into the magefile.

For example, if my magefile has the following target:

```go
func GoRuntime() {
    fmt.Printf("Go Runtime is: %s", mg.GoCmd())
}
```

If I run `mage goRuntime`, as expected, I receive the output:

```shell
Go Runtime is: go
```

However, if I run `mage -gocmd /usr/local/bin/go goRuntime`, I still receive the above output - but it should be:

```shell
Go Runtime is: /usr/local/bin/go
```

When running with `--debug` (i.e. `mage -gocmd /usr/local/bin/go --debug goRuntime`, I receive the following logs:

```shell
DEBUG: 16:31:45.427534 running magefile with mage vars:
MAGEFILE_DEBUG=1
Go Runtime is: go
```

I believe `-gocmd` should set the environment variable `MAGEFILE_GOCMD` before running the magefile in order to have `GoCmd()` function properly within the magefile.

ihgann:
Pass the `MAGEFILE_GOCMD` environment variable into the compiled binary
when running with `-gocmd` set.

Testing Done:

- Validated that running `mage -debug -gocmd <go-runtime> <target>`
  properly sets the environment variable `MAGEFILE_GOCMD=<go-runtime>`
  into the temporary binary.
- Validated that compiled magefiles will not persist the initial
  `-gocmd` when executing - though this is still toggleable by providing
  the environment variable manually.

fixes #208

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -604,6 +604,9 @@ func RunCompiled(inv Invocation, exePath string, errlog *log.Logger) int {
 	if inv.Debug {
 		c.Env = append(c.Env, "MAGEFILE_DEBUG=1")
 	}
+	if inv.GoCmd != "" {
+		c.Env = append(c.Env, fmt.Sprintf("MAGEFILE_GOCMD=%s", inv.GoCmd))
+	}
 	if inv.Timeout > 0 {
 		c.Env = append(c.Env, fmt.Sprintf("MAGEFILE_TIMEOUT=%s", inv.Timeout.String()))
 	}


natefinch:
Yeah..... the thing of it is, the MAGEFILE_GOCMD environment variable is really only meant for internal use.  It wasn't really built to be used by magefiles themselves.   I think it was a mistake to put it in the `mg` package.... but it's there now and I can't take it out, so I might as well make it consistent I guess. :)

I think this is all fine.  It definitely should not be hard coded into the compiled binary.

ihgann:
Cool - that makes sense. I'll probably switch to a different method locally (ideally if #24 ever gets implemented I can switch to args), but at least this should fix existing expectations.

natefinch:
just FYI, I've been sick for the past 8 days, but I fully intend on reviewing this as soon as my brain is back to normal.

PR #212, opened 2018-12-22
Sort target list when using mage:import

andrewkroh:
I was trying out the `mage:import` feature which is wonderful in helping reduce duplication in our monorepo. I thought it would nice to have all targets sorted when using this feature.


natefinch:
This is great, thanks!

Issue #213, opened 2018-12-23
No working directory for sh.Exec


corebreaker:
Hello,

`os/exec.Command` allow a working directory, but sometimes, we need it in `sh.Exec` too.

The easy way is to use the `exec.Command` function, but i don't like it.

I propose to pass a special variable in env parameter, like this:
```go
Exec(map[string]string { "#CWD": "/home/theUser/mydir" }, nil, nil, "my-command")
```

natefinch:
I think this is a valid concern and I think there is a way to address it. But the linked PR was not the right solution.  I think we can come up with something better and more intuitive.

philipsahli:
If have added the PR, which should solves this issue. I need to run command with mage/sh in the working directory, so I add the variable `#CWD` to `os.Getenv("PWD")`.

corebreaker:
Great !

sluedecke:
I currently use this function as a workaround:

```
func runCommand(cwd string, command string, args ...string) (ran bool, err error) {
	wd, _ := os.Getwd()
	err = os.Chdir(cwd)
	if err != nil {
		return false, err
	}
	ran, err = sh.Exec(map[string]string{"#CWD": cwd}, os.Stdout, os.Stderr, command, args...)
	os.Chdir(wd) // can return an error, but should never happen.  Smart coders might want to handle it anyways
	return ran, err
}
```

jcchavezs:
I think sh.Exec should have a way to pass the directory. For example, in Coraza we run `go test ./...` in `/my-folder` and then `go test ./...` in `/my-folder/my-subfolder` (see https://github.com/corazawaf/coraza/pull/422/files#diff-520109d035a196d29d0a86e9c4e6c98e98abf056141b84df68c0d48167b87b25R83) and there is no way to pass a global working directory.
cc @anuraaga as he usually have good ideas on how to overcome API challenges.

hamzaelsaawy:
Is there any consensus on this? 
I created #444 to create a struct to allow setting the different options in a struct, but I can update it to use options instead, as mentioned in #307.

ivanduka:
> > I currently use this function as a workaround:
> > ```
> > func runCommand(cwd string, command string, args ...string) (ran bool, err error) {
> > 	wd, _ := os.Getwd()
> > 	err = os.Chdir(cwd)
> > 	if err != nil {
> > 		return false, err
> > 	}
[...]

PR #214, opened 2018-12-23
Add a special environment variable to define the working directory

corebreaker:
Fix the issue #213


natefinch:
I don't think this is the right approach.  I don't like having magic values in the map that have special meaning and have nothing to do with environment variables.   Using os/exec would be better... we can think about extending the `sh` package... but at some point the number of options becomes too great and it just gets confusing.  I'll have to think about how to make the UX more configurable.

corebreaker:
Thanks for the response.
Ok, i understand, it'd better to use the os/exec package, so the mage/sh is useless, cause pass the working directory is really a used very often (for example to call tools like npm, angular, sass, django or git which need to be launched in another directory).
It'd not be confusing to add a new function which use structure as option instead of function arguments.

Issue #216, opened 2018-12-28
target: expanding Env and adding Glob?


kentquirk:
I would like to write some rules that are the equivalent of this Makefile target:

`cmd/chasm/chasm: $(CHAINCODEPKG)/vm/opcodes.go cmd/chasm/*.go`

The target package doesn't expand environment variables, nor does it have a way to expand globs. Obviously, I can (and did) code around these limitations, but I'm wondering if you'd be interested in a PR that does the following two things:

* Adds a new function `target.Glob`, as follows:

```go
// Glob expands each of the globs into individual sources and reports if any of the 
// resulting sources have been modified more recently
// than the destination. Glob does not descend into directories, it literally
// just checks the modtime of each source. If the destination
// file doesn't exist, it always returns true and nil. It is an error for a glob
// to return an empty result.
func Glob(dst string, globs ...string) (bool, error) {
```

* Expands envvars in source and destination of all the target functions: `Path`, `Dir`, and `Glob`.

This one is slightly trickier because then it might also imply that we should also add `PathWith`, `DirWith`, and `GlobWith` to be consistent with `sh`. 

I'd be willing to do that if you felt it was important. I guess the question there is whether the "With" approach is really the right strategy -- does it provide anything that can't instead be handled with `os.Setenv`?

natefinch:
Yeah I agree that the target package needs some love.  I think adding auto expanding of environment variables and auto globbing would be a good addition.  

I actually think these could get added to the default functions.  I'll have to think more about exactly how it would elaffect usage.

kentquirk:
I currently have the Glob function described above that basically just passes the result of the Glob on to target.Path.

But you're right -- both features could be incorporated into Path and Dir without breaking existing uses (except for that 1 person in a million who foolishly has filenames that incorporate characters the shell uses, like `*`, `{`, `}`, and `$`). 

I'll try putting together a PR that does this, and promise I won't be offended if you don't like it or want a different approach.

natefinch:
Thank you!  PRs most appreciated :)

natefinch:
Merged... I guess this can be closed :)

PR #219, opened 2018-12-31
Add expansion of envvars and a Glob function to target package

kentquirk:
Addresses #216.

Expands envvars in Path and Dir functions, and adds a Glob function.

Adds a bunch of tests for the new functionality.


kentquirk:
@natefinch Just a bump in case you didn't see this.

Issue #211, opened 2018-12-20
Improve sh.Copy


ntrrg:
Hello :smile: I want to improve `sh.Copy` adding support for multiple files and directories, but I was wondering:

* For multiple files, the signature of `sh.Copy` will be `func Copy(dst string, src ...string) error`, so it should work as always but break code like `func MyFunction(arg string, copyFunc func(string, string) error) error`, is this important? obviously this might be easily fixed with `func MultiCopy(dst string, src ...string) error`, but that means one extra exported function

* For directories, is there any reason why it doesn't support directories? I mean, is that intentional? or supporting directories would be fine?

ntrrg:
So, this could be the best way to do this?

* `func Copy(dst, src string) error`
* `func MultiCopy(dst string, src ...string) error`
* `func CopyDir(dst, src string) error`
* `func MultiCopyDir(dst string, src ...string) error`

PR #220, opened 2019-01-06
Add copying helpers

ntrrg:
Since the sh.Copy helper just support one file operations, copying directories or multiple file may require help from the OS, which brings compatibility issues (one of the main reasons why we don't use Make). This PR aims to add the following helpers:

* `sh.CopyDir`: copies source directory content to the given destination. This will fail if destination doesn't exist.
* `sh.CopyDirAll`: same as above, but creates the destination directory if it doesn't exist.
* `sh.CopyAny`: copies the source elements to the given destination. Destination must be a directory.

Resolves: #211

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,11 @@ mage_output_file.go
 # Goland
 .idea
 *.iml
+
+# Vim
+*.sw[op]
+*.vim
+
+# GNU Screen
+.screenrc
+
--- a/sh/helpers.go
+++ b/sh/helpers.go
@@ -87,18 +87,31 @@ func CopyDir(dst, src string) error {
 // are:
 //
 // * (FAIL) No source.
+//
 // * (FAIL) Non existent source.
+//
 // * (PASS) File to non existent file.
+//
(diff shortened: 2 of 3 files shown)


natefinch:
So, I was looking at this again, and I think copying the `cp` command is maybe not the best way to do this. I can't tell you how often I do the wrong thing with cp because I expect it to work one way and it works some other way.  The fact that the comment on the Cp function is as long as the function is a red flag, IMO.

I'd prefer multiple functions that were more obvious in how they work and explicitly fail if you use them incorrectly.

CopyContents - copies all the files and directories in an existing directory into another existing directory.
CopyDir - creates a new directory all files and directories in an existing directory into it (basically just mkdirall and then CopyContents)

And maybe CopyTo - takes a list of files and copies them to an existing directory.

ntrrg:
Yes, actually it could be hard to follow the code some times :sweat_smile: I will try your approach

ntrrg:
If you can, review the code again :) I will be working on the tests.

ntrrg:
I think this may be implemented by users that really need it haha so I am closing the PR :sweat_smile:

Issue #240, opened 2019-05-11
"Invalid type for dependent function" does not have any context


misha-tectonic:
If an invalid kind of dependency is passed into `mg.Deps`, then it's pretty hard to nail down what has happened, as the error message does not provide any location where that happened:

`Error: Invalid type for dependent function: func(string, string). Dependencies must be func(), func() error, func(context.Context), func(context.Context) error, or the same method on an mg.Namespace.`

It would be nice to be able to point to at least `mg.Deps` invocation, or, better, to the argument that caused the problem.

PR #241, opened 2019-05-11
Improve error message on wrong type of mg.Deps argument (#240)

misha-tectonic:
This change adds a location of mg.Deps invocation, shrinking the search space
from "all your build scripts" to "this particular mg.Deps invocation"

### DIFF
--- a/mage/main_test.go
+++ b/mage/main_test.go
@@ -1261,6 +1261,25 @@ func TestAliasToImport(t *testing.T) {
 
 }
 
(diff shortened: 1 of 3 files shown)


natefinch:
Looks good, but can you write a test for it?  Can be pretty simple, just something that'll break if we change the number of function calls between the name and the error?

misha-tectonic:
Sure, added a test.

natefinch:
hmm probably some problem on tip from go modules, I'll look into it.

misha-tectonic:
Yes, I saw this test failing with modules.

PR #246, opened 2019-06-07
set working directory on Cmd if variable #CWD is set (#213)

philipsahli:
This allows to specify a directory (the value of a variable `#CWD` in the env list) where the commands are executed. Useful when the Magefile is in a different directory than where you want to execute the cmd's.


philipsahli:
Check on continuous-integration/travis-ci does not finish. I will put the PR still to review.

natefinch:
This is a good idea. I think we've figured out the CI problems in another PR.

natefinch:
I think this will be fixed by #277 with a more consistent approach with how the rest of mage works.

Issue #224, opened 2019-01-29
TestMixedMageImports fail on macOS in modules mode (current master)


misha-tectonic:
Hello.

Running `go test ./...` on macOS, commit 5dee1ad6efb4c25f634bda90d11fb5157c80e5e8 gives the following:

    --- FAIL: TestMixedMageImports (0.17s)
        main_test.go:249: expected to exit with code 0, but got 1, stderr: build command-line-arguments: cannot find module for path _/Users/dottedmag/srcs/mage/mage/testdata/mixed_lib_files/subdir
            Error: error compiling magefiles
        main_test.go:254: expected "Targets:\n  build    \n" but got ""

natefinch:
Sorry, missed this when it was first filed.  It works for me on my mac, but I'm not using modules, that may be the problem.

misha-tectonic:
Right, the test passes if I run it from `$GOPATH`

nlowe:
Note this doesn't seem to be mac specific. This also is causing failures on travis with `Go: tip` and locally with 1.12:

```bash
$ docker run -it --rm -v "$(pwd):/mage" -w /mage golang:1.12 go test -v -run TestMixedMageImports ./mage/...                                                                     === RUN   TestMixedMageImports                                                                                    --- FAIL: TestMixedMageImports (0.12s)
    main_test.go:249: expected to exit with code 0, but got 1, stderr: build command-line-arguments: cannot find module for path _/mage/mage/testdata/mixed_lib_files/subdir
        Error: error compiling magefiles
    main_test.go:254: expected "Targets:\n  build    \n" but got ""
FAIL
FAIL    github.com/magefile/mage/mage   0.132s
```

nlowe:
So you can [sort of work around this by adding `go_import_path` to the `.travis.yml` file](https://travis-ci.org/nlowe/mage/builds/549449941), but something seems to have changed from 1.12 in dev that breaks it.

nlowe:
Ah, so the definition of `GO111MODULE` changed in 1.13 (golang/go#31857). It's treated as if it were set to `auto` when unset. In 1.13+, this means if `go` is run in a directory with a `go.mod` file, or in a subdirectory of a directory that contains a `go.mod` file, the go modules behavior is enabled.

[I was able to work around this](https://travis-ci.org/nlowe/mage/builds/549453398) with this patch:

```diff
diff --git a/.travis.yml b/.travis.yml
index f7e0703..895ed9b 100644
[...]

PR #251, opened 2019-06-23
New Flag: -nodeps: skip target dependencies when executing target

### LINKED ISSUE #250: Support "Exclusive" Mode to execute a task without dependencies
[Cake](https://github.com/cake-build/cake) (a similar tool for writing build scripts in C#) supports a flag called `--exclusive`. This lets Cake execute a specific task without executing its dependencies. This allows you to quickly re-test a specific task (for example, re-build a Docker container without re-running tests if you just changed the dockerfile) and is particularly well suited for build environments like Jenkins where you can separate tasks for visibility.

In most of the Cake projects we use, we wire up dependencies normally, so if a developer runs the `Docker` test, they restore packages, build code, and run tests. In our Jenkins file, we split each task to a separate stage (called with `--exclusive`) in Jenkins so you get nice timing information and better visibility in Blue Ocean.

Is this something Mage could potentially support? I'd be happy to look into sending a PR.

nlowe:
When you run mage with '-nodeps', mage will skip any Deps/SerialDeps
in the target being executed. This is useful when you want to run a
single target or need to run targets with dependencies locally but
without dependencies in CI for timing / visibility reasons.

Fixes #250

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -100,6 +100,7 @@ type Invocation struct {
 	List       bool          // tells the magefile to print out a list of targets
 	Help       bool          // tells the magefile to print out help for a specific target
 	Keep       bool          // tells mage to keep the generated main file after compiling
+	NoDeps     bool          // tells mage to execute the given target without dependencies
 	Timeout    time.Duration // tells mage to set a timeout to running the targets
 	CompileOut string        // tells mage to compile a static binary to this path, but not execute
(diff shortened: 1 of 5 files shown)


nlowe:
Hmm, I'm not sure what that failure is on travis for `Go: tip`. I can't repro locally with 1.12. Did go modules get flipped on by default?

nlowe:
Ah, we're hitting #224.

PR #257, opened 2019-07-18
fix tests when running in module mode

natefinch:
Turns out that relative imports in module mode don't work well.  Who knew?  Is it a bug in the compiler? Maybe? Either way, this fixes it.


natefinch:
See https://github.com/golang/go/issues/33163

PR #261, opened 2019-08-15
Add RunVWith command

### LINKED ISSUE #260: Add RunVWith
We have `mg.RunV`, it would be nice to also have `mg.RunVWith`. I've tried setting `MAGE_VERBOSE` in the map I pass into `mg.RunWith` but it doesn't seem to work.

eberkund:
Resolves #260

### DIFF
--- a/sh/cmd.go
+++ b/sh/cmd.go
@@ -69,6 +69,12 @@ func RunWith(env map[string]string, cmd string, args ...string) error {
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


eberkund:
The failing CI is due to an unrelated issue I believe.

natefinch:
Well that's interesting. It's probably a bug in tip of some sort.  I'll take a look.

crazy-max:
@natefinch Really need this feature :sweat_smile: Any ETA?

PR #268, opened 2019-10-09
Move panic handling into run() for simpler code

misha-tectonic:
This also corrects the following problem:

- several tasks are waiting on a single subtask
- subtask panics
- first task receives a panic and handles it
- other tasks receive nil, as t.err is not set


misha-tectonic:
My apologies, wrong repo.

PR #271, opened 2019-11-01
Use extended globbing syntax

WIZARDISHUNGRY:
Hi - I don't actually intended for this version of mage to be merged (since it adds a dependency). But I would like it if we supported double star expressions in Glob.

See https://github.com/golang/go/issues/11862 
https://github.com/bmatcuk/doublestar

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/magefile/mage
 
 go 1.12
+
+require github.com/bmatcuk/doublestar v1.1.5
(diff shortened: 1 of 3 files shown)


natefinch:
So, right now, Mage has no external dependencies.  I'd like to keep it that way unless the benefit is really large.  This change could be done in a personal repo, or maybe in a mage-contrib repo that has a less restrictive import policy.

PR #273, opened 2019-11-08
add support for concurrent job limits

### LINKED ISSUE #38: Add helper for parallelisation limits
The code is trivial, but it might be nice to add a helper to provide parallelisation limits.

kortschak:
Please take a look.

Fixes #38.

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -102,6 +102,7 @@ type Invocation struct {
 	Help       bool          // tells the magefile to print out help for a specific target
 	Keep       bool          // tells mage to keep the generated main file after compiling
 	Timeout    time.Duration // tells mage to set a timeout to running the targets
+	MaxProcs   int           // tells mage to set a maximum number of concurrent dependencies
 	CompileOut string        // tells mage to compile a static binary to this path, but not execute
 	GOOS       string        // sets the GOOS when producing a binary with -compileout
 	GOARCH     string        // sets the GOARCH when producing a binary with -compileout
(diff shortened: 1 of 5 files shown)


natefinch:
The problem with this is that it can really easily deadlock.  To run dependencies last to first, mage just fires off a bunch of goroutines and waits for them all to exit. It will do this several levels deep, so you'll often have N goroutines firing off. So if you have T1->T2->T3 and j is 1, T2 will use up a slot and then it'll deadlock then it wants to start T3 but there's no more slots.

I don't really think this is a huge problem. This is limited by how deep your dependency tree is, as defined by calling mg.Deps, and it's almost never going to be more than 3 or 4 levels deep, so, what, like maybe max 10 goroutines waiting?

kortschak:
The use case that I'm interested in is when mage is used to control workflows with potentially very resource expensive parts where these are run in parallel and because of the expense can bring down the workflow if too many are run at once. Maybe this needs to be a component that is not so tightly integrated into the mage runtime?

natefinch:
You can use mg.SerialDeps to run dependencies without parallelization. Maybe that is sufficient?


### func SerialDeps(fns ...interface{})

> SerialDeps is like Deps except it runs each dependency serially, instead of in parallel. This can be useful for resource intensive dependencies that shouldn't be run at the same time.

kortschak:
No, not really. I'm look at this as a workflow mananager for bioinformatic pipelines. The workflow components are large and long lived, the machines are also large. So often you can run some parts in parallel, but the parallelisation is limited and the constraint is often system memory. The two approaches (out of the box) either risk having components brought down by the oom killer (unlimited parallelisation) or making jobs take longer and wasting computing resources (the sequential approach). It would be nice if there were a middle way.

natefinch:
Can you give me a magefile that stubs out some targets and how they interact, so I have a better idea of what you need?

As an aside, it's possible the answer is just not to use mage targets directly. Clearly this is something that can be modeled in Go, so the answer may be to write it directly for your use case, and use mage to initiate it (or not... maybe mage is not the right tool).

That being said, I do think it's valuable to be able to easily control the parallelism of running tasks, so I'll try to think about how dependencies are run and if there's a way to limit them without easily getting deadlocked.

ghostsquad:
> The problem with this is that it can really easily deadlock. To run dependencies last to first, mage just fires off a bunch of goroutines and waits for them all to exit. It will do this several levels deep, so you'll often have N goroutines firing off. So if you have T1->T2->T3 and j is 1, T2 will use up a slot and then it'll deadlock when it wants to start T3 but there's no more slots.

I don't see how this can't be done, since if you set parallelism to `1`, it should basically just turn into a serial runner.

If the dependencies were resolved like a graph:

https://dnaeon.github.io/dependency-graph-resolution-algorithm-in-go/

This would not be a problem.

Issue #243, opened 2019-05-17
Magefiles dependencies in go.mod


ntrrg:
Hi, if you try to import any library using Mage, you will download Mage as a dependency. You can try that with Hugo:

`main.go`

```go
package main

import (
	"fmt"

	"github.com/gohugoio/hugo/livereload"
)

func main() {
	fmt.Printf("%s\n", livereload.Handler)
}
```

```shell-session
$ go mod init go-test
$ go run main.go
```

The reason is because the `go.mod` file from Hugo includes Mage and any other magefile dependency in it

natefinch:
Yes. That's how dependencies work. Mage is a dependency of hugo's code. If you run `go mod tidy` it might clean up the dependencies you don't actually need from hugo's transitive dependencies, but I'm not 100% sure on that. 

This is similar to when you import a library that uses 3rd party test libraries... those get put into your dependencies as well. This is basically unavoidable.

ntrrg:
Yes, I know that this is not actually a Mage issue, but we could avoid it providing something like a `-D` flag for setting the working directory of the compiled magefiles, so you could put all your magefiles in other directory (`.mage` for example) and run mage as `mage -d .mage -D .`, that way the `go.mod` file doesn't get populated with magefiles dependencies

natefinch:
So, I think this is partly a bug of -d that it sets the working directory to the -d directory. I had intended it to take the directory you're in as the working directory.  I wonder if it's too late to make -d work that way...   but separating that out might be better.

ntrrg:
I was thinking the same, but there could be someone relying on that behavior, so I wasn't sure how important is keeping it in that way. I did tried getting this working and just some lines were needed

PR #277, opened 2019-11-29
Set magefiles working directory

ntrrg:
This adds a `-w` flag for changing the magefiles working directory. Currently reading magefiles from other directory with `-d` flag set it as working directory for the magefiles.

This doesn't break any existing code since the `-w` flag default value is the same as `-d`, just people using explicitly a `-w` flag will get this behavior.

Fixes: #243

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,11 @@ mage_output_file.go
 # Goland
 .idea
 *.iml
+
+# Vim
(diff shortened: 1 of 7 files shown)


ntrrg:
Yes, tar and Git :sweat_smile: great, I will work on that :)

PR #278, opened 2020-01-05
Add ability to defer binary template to input parameter

fubarhouse:
I am trying to achieve the ability for some customisation which are physically gated by upstream decisions. In order to facilitate more changes downstream it makes sense that the template can be switched out for a user-provided template to prevent the conversation or change from happening upstream.

This PR reflects this change, to provide a new input parameter which allows for the user to provide a path to the custom template. It appears to be working well, so please feel free to give feedback.


natefinch:
This is going to be very brittle. Any change to the upstream code or template will likely break a customized template, and keeping backwards compatibility is important to me.  It also would seem to require really in-depth knowledge of the mage codebase, which I doubt very many people will want to learn.

PR #293, opened 2020-03-03
update home page with a better example

guumaster:
It was confusing for me to get started with the current example, I don't have `dep` installed and I think the example with `go mod download` may be a better example.

Also change the order of installation methods, recommending `go modules` instead of $GOPATH.


guumaster:
@natefinch do you need something else on this branch before merging it?

natefinch:
Sorry, thought I already merged this. Will do so now.

PR #298, opened 2020-03-31
Improve error message for failed imports

aknuds1:
This PR improves the error message when importing packages fails, for example because an imported package has a build tag (e.g. `mage`). A test is also included.

Plus, I edited the documentation to clarify how Mage doesn't support importing package files with build tags.

### DIFF
--- a/mage/import_test.go
+++ b/mage/import_test.go
@@ -187,3 +187,31 @@ func TestMageImportsOneLine(t *testing.T) {
 		t.Fatalf("expected: %q got: %q", expected, actual)
 	}
 }
+
+func TestMageImportsTaggedPackage(t *testing.T) {
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	inv := Invocation{
+		Dir:    "./testdata/mageimport/tagged",
+		Stdout: stdout,
+		Stderr: stderr,
+		List:   true,
+	}
+
+	code := Invoke(inv)
+	if code != 1 {
+		t.Fatalf("expected to exit with code 1, but got %v, stdout:\n%s\nstderr:\n%s", code, stdout, stderr)
+	}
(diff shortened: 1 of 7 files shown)


natefinch:
As mentioned in slack, adding the mage build tag would change the behavior for folks currently using mage, and that could break stuff. You don't need the mage tag on stuff you import.  As you pointed out in slack, the comment on this in the docs is not clear at all, and I welcome proposed changes to the wording.

I do like the fix to print a better error message, however.

aknuds1:
@natefinch Thanks, I modified the PR to only improve the error message (along with a test). Please have another look.

mage/import_test.go
```diff
@@ -187,3 +187,31 @@ func TestMageImportsOneLine(t *testing.T) {
...
+
+	code := Invoke(inv)
+	if code != 1 {
+		t.Fatalf("expected to exit with code 1, but got %v, stdout:\n%s\nstderr:\n%s", code, stdout, stderr)
+	}
+
+	actual := stderr.String()
+	expected := `
+Error parsing magefiles: error running "go list -f {{.Dir}}||{{.Name}} github.com/magefile/mage/mage/testdata/mageimport/tagged/pkg": exit status 1
+can't load package: package github.com/magefile/mage/mage/testdata/mageimport/tagged/pkg: build constraints exclude all Go files in /home/arve/Projects/mage/mage/testdata/mageimport/tagged/pkg
```
natefinch, line 209: Oh yeah, this test won't work as is... you could just look for a prefix of the string up to "exclude all Go files in" ... I don't think we need to worry that the path will be wrong.

aknuds1: @natefinch Ooops, I see the problem now :man_facepalming: Thanks, I will fix it.

Issue #248, opened 2019-06-18
when using a compiled mage binary, the verbose flag isn't propagated


Oneymus:
When using a pre-compiled static `mage` binary, the `-v` flag does not produce the proper output.

Using this simple example:
```go
func ShowVerboseDep() error {
	sh.Run("echo", "hello, mage!")
	return nil
}

func ShowVerbose() error {
	mg.Deps(ShowVerboseDep)
	sh.Run("echo", "goodbye, mage!")
	return nil
}
```

Output:
```bash
$ mage -v showverbose
Running target: ShowVerbose
Running dependency: ShowVerboseDep
exec: echo hello, mage!
hello, mage!
exec: echo goodbye, mage!
goodbye, mage!
```

```bash
$ ./mage -v showverbose
Running target: ShowVerbose
exec: echo hello, mage!
exec: echo goodbye, mage!
```

We're missing both the shell output and the dependency log lines.

When using the `mage` command directly, `mage` sets the `MAGEFILE_VERBOSE` environment variable based on the `-v` flag (see [here](https://github.com/magefile/mage/blob/master/mage/main.go#L611)).

However, the generated `mage_output_file.go` from which the compiled binary is built does no such thing.

This is relevant because both `sh` and `mg` inspect the `MAGEFILE_VERBOSE` environment variable (see [here](https://github.com/magefile/mage/blob/master/mg/runtime.go#L30-L34)); there doesn't appear to be any other way to signal to these libraries that we want verbose output.

I propose we update the generated `mage_output_file.go` to set the `MAGEFILE_VERBOSE` environment variable just like we do for the `mage` command.

philipsahli:
I observed the same issue some days ago. My workaround is to add the variable in an init function, because at the moment I feel that in any case the output of the commands are interesting: 

    // When using a static binary (by doing `mage -o gob`, the arg `-v` seems to be ignored.
    os.Setenv("MAGEFILE_VERBOSE", "1")

PR #300, opened 2020-04-15
Ensure MAGEFILE_VERBOSE set in compiled magefile

axw:
When using `mage -compile=foo`, ensure that passing
`-v` to the produced binary sets MAGEFILE_VERBOSE.
This is necessary for `mg.Verbose()` to return the
correct value.

Fixes https://github.com/magefile/mage/issues/248

### DIFF
--- a/mage/testdata/compiled/custom.go
+++ b/mage/testdata/compiled/custom.go
@@ -14,7 +14,9 @@ var Default = Deploy
 
 // This is very verbose.
 func TestVerbose() {
-	log.Println("hi!")
+	if mg.Verbose() {
+		log.Println("hi!")
+	}
 }
 
 // This is the synopsis for Deploy. This part shouldn't show up.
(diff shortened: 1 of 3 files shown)


mage/testdata/compiled/custom.go
```diff
@@ -14,7 +14,9 @@ var Default = Deploy
 
 // This is very verbose.
 func TestVerbose() {
-	log.Println("hi!")
+	if mg.Verbose() {
+		log.Println("hi!")
+	}
```
natefinch, line 19: log's output is already determined by the flag, so I think this needs to be a separate test, so that we're not munging the two tests.  One should test that log does the right thing and one should test that mg.Verbose does the right thing.

axw: Done

axw:
I've reverted the change to TestVerbose, and introduced a new target and test.

Issue #285, opened 2020-01-23
Add color output for `mage -l` targets


mirogta:
We now have ~30 mage build targets so the output from `mage -l` isn't very easy on eyes.

It would be nicer by coloring the build targets with some arbitrary color (cyan?) while leaving the description (comments on public mage funcs) in the default color.

Inspired by https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

natefinch:
Hmm.. I'm not against this, but I always worry about terminal compatibility.

mirogta:
Fair point... Is there a way to find out all people who are currently using or looking for using Mage, and then ask them if they use a color terminal?

Would it be compatible with your terminal?

ghostsquad:
you could look at the `TERM` env var. If it's set to something like `xterm-256color` then enable by default. You can also provide options such as

```
--color
--no-color
```

to force one behavior regardless.

https://www.gnu.org/software/gettext/manual/html_node/The-TERM-variable.html

mirogta:
@natefinch I have a working code and a test, but not quite ready yet for a PR, because...

I have a passing test, similar to the existing `TestMageImportsList` in `import_test.go` but instead of the **expected** string with a simple string which looks like this:

```
	actual := stdout.String()
	expected := `
Targets:
  buildSubdir        Builds stuff.
  ns:deploy          deploys stuff.
  root               
  zz:buildSubdir2    Builds stuff.
  zz:ns:deploy2*     deploys stuff.

* default target
`[1:]

	if actual != expected {
		t.Logf("expected: %q", expected)
		t.Logf("  actual: %q", actual)
		t.Fatalf("expected:\n%v\n\ngot:\n%v", expected, actual)
	}
```

... I have the **expected** string with the ANSI colored text:

```
	actual := stdout.String()
	expected := `
Targets:
  \033[36mbuildSubdir%s\033[0m        Builds stuff.
  \033[36mns:deploy%s\033[0m          deploys stuff.
  \033[36mroot%s\033[0m               
  \033[36mzz:buildSubdir2%s\033[0m    Builds stuff.
  \033[36mzz:ns:deploy2*%s\033[0m     deploys stuff.

* default target
`[1:]

	if actual != expected {
		t.Logf("expected: %q", expected)
		t.Logf("  actual: %q", actual)
		t.Fatalf("expected:\n%v\n\ngot:\n%v", expected, actual)
	}

```

[...]

PR #301, opened 2020-04-25
#285 add color output for mage list

mirogta:
I've chosen a cyan color, because it's a good contrast on a typically black terminal (better than e.g. blue) and it's neutral, as opposed to e.g. green, red, yellow.

Added one test to verify the colorized output. All other tests remain without the colorized output, because I unset the TERM env var, using `resetTerm` func. I've added a call to the `resetTerm` into the `testmain` func, but it wasn't enough and some tests were still failing so I've added an explicit call to the `resetTerm` to them as well.


mage/template.go
```diff
@@ -93,6 +93,60 @@ Options:
...
+
+	// terminalSupportsColor checks if the current console supports color output
+	//
+	// Supported:
+	// 	linux, mac, or windows's ConEmu, Cmder, putty, git-bash.exe
+	// Not supported:
+	// 	windows cmd.exe, powerShell.exe
+	terminalSupportsColor := func() bool {
+		envTerm := os.Getenv("TERM")
+		if strings.Contains(envTerm, "xterm") {
```
natefinch, line 119: This kinda skeeves me out.  Are we sure there are no existing terminals that include the substring xterm that don't support color? (I ask because I have no clue)

natefinch:
Sorry this sat around so long, I can't believe it's been 23 days.  Time no longer has meaning, I guess.

I love that you did this without requiring an import of a 3rd party library.

mirogta:
> Sorry this sat around so long, I can't believe it's been 23 days. Time no longer has meaning, I guess.
> 
> I love that you did this without requiring an import of a 3rd party library.

No problem at all, my time is also completely skewed - can't believe it's almost June! I'll crack on the suggestions :-) Thank you

mage/template.go
mirogta: Yeah, the best I could find was that there's "xterm-mono". I've added a condition to disable terminal support of this is detected.

See https://github.com/rust-lang/cargo/issues/7646

mirogta:
@natefinch please have a look if it looks sensible and I'll add more tests around it. I know, TDD and so on, but it would be quite difficult to follow in this case while I was just exploring the best approach.

Please note I've left the new `color.go` file with a test there, even though it's not really used in any code (yet). It has helped me massively to get the code working, which I could then copy&paste with just minor modifications into the `template.go`.

mage/template.go
natefinch: Sorry, I missed this was a question for me. Yes, let's do that.

mirogta: Cool - done, see the update.

I've used a map rather than a list for performance, so that I don't need to iterate over it and just use the terminal name as a key.

mirogta: "It works on my machine" including all tests, but having some trouble with tests in Travis CI. Trying to find out what is the TERM set to...

mirogta: Looking good now. Ready for a merge & release?

mage/main_test.go
```diff
@@ -57,9 +57,31 @@ func testmain(m *testing.M) int {
...
+	}
+	if err := os.Unsetenv(mg.TargetColorEnv); err != nil {
+		log.Fatal(err)
+	}
+	resetTerm()
 	return m.Run()
 }
 
+func resetTerm() {
+	if term, exists := os.LookupEnv("TERM"); exists == true {
```
natefinch, line 74:
```suggestion
	if term, exists := os.LookupEnv("TERM"); exists {
```

mg/runtime.go
```diff
@@ -85,5 +115,22 @@ func CacheDir() string {
...
+func EnableColor() bool {
+	b, _ := strconv.ParseBool(os.Getenv(EnableColorEnv))
+	return b
+}
+
+// TargetColor returns the configured ANSI color name a color output.
+func TargetColor() string {
+	s, exists := os.LookupEnv(TargetColorEnv)
+	if exists == true {
+		if c, ok := getAnsiColor(s); ok == true {
```
natefinch, line 128:
```suggestion
		if c, ok := getAnsiColor(s); ok {
```

mg/runtime.go
```diff
@@ -85,5 +115,22 @@ func CacheDir() string {
...
+// EnableColor reports whether the user has requested to enable a color output.
+func EnableColor() bool {
+	b, _ := strconv.ParseBool(os.Getenv(EnableColorEnv))
+	return b
+}
+
+// TargetColor returns the configured ANSI color name a color output.
+func TargetColor() string {
+	s, exists := os.LookupEnv(TargetColorEnv)
+	if exists == true {
```
natefinch, line 127:
```suggestion
	if exists {
```

Issue #304, opened 2020-05-06
sh.Run and sh.Output sh.OutCmd fail on a go build command but running it directly on local machine works


varunturlapati:
I setup my mage target to build locally and it worked fine. Then I wanted to introduce ldflags to get build info embedded as well. The build target errors out.

Expected:
Build succeeds

Actual:
```
go: finding github.myorg.com/myrepo/heyo/version.BuildSha=4accbfe latest
go: finding github.myorg.com/myrepo/heyo/version.ReleaseType=localdev' latest
go: finding github.myorg.com/myrepo/heyo/version.BuildVersion=local latest
can't load package: package github.myorg.com/myrepo/heyo/version.BuildDate=20200505: no matching versions for query "latest"
can't load package: package -X: malformed module path "-X": leading dash
can't load package: package github.myorg.com/myrepo/heyo/version.BuildSha=4accbfe: no matching versions for query "latest"
can't load package: package github.myorg.com/myrepo/heyo/version.BuildVersion=local: no matching versions for query "latest"
can't load package: package github.myorg.com/myrepo/heyo/version.ReleaseType=localdev': no matching versions for query "latest"
can't load package: package -os: malformed module path "-os": leading dash
can't load package: package darwin: malformed module path "darwin": missing dot in first path element
can't load package: package -arch: malformed module path "-arch": leading dash
can't load package: package amd64: malformed module path "amd64": missing dot in first path element
can't load package: package -output: malformed module path "-output": leading dash
can't load package: package targets/ocs_heyo_darwin_amd64: malformed module path "targets/ocs_heyo_darwin_amd64": missing dot in first path element

Error: running "gox -ldflags '-X github.myorg.com/myrepo/heyo/version.BuildDate=20200505 -X github.myorg.com/myrepo/heyo/version.BuildSha=4accbfe -X github.myorg.com/myrepo/heyo/version.BuildVersion=local -X github.myorg.com/myrepo/heyo/version.ReleaseType=localdev' -os darwin -arch amd64 -output targets/ocs_heyo_darwin_amd64 github.myorg.com/myrepo/heyo/cmd" failed with exit code 1
```

Then as an experiment I simply copied the whole command that it says it ran in the last line.

[...]

PR #306, opened 2020-05-12
sh.run(): quoted strings before join

shouldsee:
Unquoted string may contain whitespace, 
after Joining those strings, the result cannot be split back to recover the original call.
```go 
ret:= sh.RunWith(nil, "go", "build", "-ldflags", "-X /common/hugo.commitHash= -X /common/hugo.buildDate=", "hugo")
```
see #304


sh/cmd.go
```diff
@@ -132,11 +132,14 @@ func run(env map[string]string, stdout, stderr io.Writer, cmd string, args ...st
 	c.Stderr = stderr
 	c.Stdout = stdout
 	c.Stdin = os.Stdin
-	log.Println("exec:", cmd, strings.Join(args, " "))
+
+	var quoted []string; for i:=0; i<len(args); i++ { 
+		quoted=append(quoted,fmt.Sprintf("\"%s\"",args[i]));
```
natefinch, line 137: if you use %q, it'll escape quotes in the string, which might make debugging easier. It won't make a perfect reusable command line, but neither would this, if a string happened to contain quotes.
```suggestion
		quoted=append(quoted,fmt.Sprintf("%q",args[i]));
```

sh/cmd.go
```diff
@@ -132,11 +132,14 @@ func run(env map[string]string, stdout, stderr io.Writer, cmd string, args ...st
 	c.Stderr = stderr
 	c.Stdout = stdout
 	c.Stdin = os.Stdin
-	log.Println("exec:", cmd, strings.Join(args, " "))
+
+	var quoted []string; for i:=0; i<len(args); i++ {
```
natefinch, line 136: this doesn't look properly go-formatted.  Also, you can just do a range over args to get the indices.
```suggestion
	var quoted []string 
	for i := range args {
```

sh/cmd.go
```diff
@@ -132,11 +132,14 @@ func run(env map[string]string, stdout, stderr io.Writer, cmd string, args ...st
 	c.Stderr = stderr
 	c.Stdout = stdout
 	c.Stdin = os.Stdin
-	log.Println("exec:", cmd, strings.Join(args, " "))
+
+	var quoted []string; for i:=0; i<len(args); i++ { 
+		quoted=append(quoted,fmt.Sprintf("\"%s\"",args[i]));
+	}
+	log.Println("[magefile-exec]:", cmd,  strings.Join(quoted, " "))
```
natefinch, line 139: for now, can you change this back to just "exec:" ?  I can understand the desire to change it to be more obvious, but I am hesitant to change any output that someone might depend on, without good reason.
```suggestion
	log.Println("exec:", cmd,  strings.Join(quoted, " "))
```

sh/cmd.go
shouldsee: yes I was just starting to learn go after making this PR :D

sh/cmd.go
shouldsee:
```go
package main
import "fmt"


func main(){
	args := []string{"something",`"not something"`, `'fbar'`}
	for _,arg := range args{
		fmt.Printf("%q\n",arg)

	}

}
```

```
"something"
"\"not something\""
"'fbar'"
```

shouldsee: Nice use of formatting!

sh/cmd.go
```diff
@@ -132,11 +132,15 @@ func run(env map[string]string, stdout, stderr io.Writer, cmd string, args ...st
 	c.Stderr = stderr
 	c.Stdout = stdout
 	c.Stdin = os.Stdin
-	log.Println("exec:", cmd, strings.Join(args, " "))
+
+	var quoted []string 
+	for i := range args {
+		quoted=append(quoted,fmt.Sprintf("%q",args[i]));
```
natefinch, line 138:
```suggestion
		quoted = append(quoted, fmt.Sprintf("%q", args[i]));
```

natefinch (approved):

Issue #291, opened 2020-02-11
sh: Allow disabling of `run` log call


cflewis:
https://github.com/magefile/mage/blob/v1.9.0/sh/cmd.go#L129 executes a log statement that cannot be overridden. The only means of stopping this would be to call `log.SetOutput()` and modify this globally.

It would be nice if `sh/` would allow setting a logger (even as a package global would be fine IMHO) so the logging could be controlled/intercepted.

natefinch:
Hmm, yeah, this is a good point.

ghostsquad:
@natefinch how would you feel about just removing this line for now? since it potentially also is a security concern, as it might print secrets to the console via cmd args.

PR #308, opened 2020-05-20
Only log executed command in verbose mode

rvolosatovs:
Close https://github.com/magefile/mage/issues/291


natefinch:
So, this is already the case.  Perhaps it was a mistake, but mage sets up the default log.Logger to only print out when verbose is true.

Issue #266, opened 2019-09-18
How to run a cleanup function on control-c?


comicsads:
I would like to run a target/function when another target gets closed using ctrl-c. The reason for this is I'm creating a webserver using go, and have a mage target to build an executable and run it, and the only way to close it is by using ctrl-c. I would like to be able to run the target Cleanup() after hitting ctrl-c.
Is there anyway to do this? Is this what context.Context() is for?
Thanks in advance!

natefinch:
Hmm yeah, that's a feature I intended to implement with the context, but currently it's not implemented.  What *should* happen is that ctrl-c should cancel the context and then exit after a short wait.

PR #313, opened 2020-08-11
mage: cancel context on SIGINT

### LINKED ISSUE #269: Signal can't be caught by the underlying binary
Hi everyone

Moved my code from makefile to mage and I'm very pleased so thank you for the great work! 

I'm running another go binary which has a loop running and this code to catch signals
```
stopChan := make(chan os.Signal)
	signal.Notify(stopChan, syscall.SIGTERM, syscall.SIGINT, os.Interrupt, syscall.SIGQUIT, syscall.SIGHUP)
```

then this code a bit later:
```
<-stopChan
log.Info("Shutting down gracefully...")
```

but when i ctrl + c my mage running process the underlying process doesn't receive any signal.
I've seen it with other executables that do cleanup on terminal as well such as terraform.

Let me know if you have any idea on how to approach a fix here or maybe if I'm missing something.

Thanks!

pmcatominey:
On receiving an interrupt signal, mage cancels the context allowing the magefile
to perform any cleanup before exiting.

A second interrupt signal will kill the magefile process without delay.

The behaviour for a timeout remains unchanged (context is canclled and the magefile
exits).

Fixes #269, #266.


mage/main_test.go
```diff
@@ -1235,7 +1236,7 @@ func TestCompiledEnvironmentVars(t *testing.T) {
 	if err == nil {
 		t.Fatalf("expected an error because of timeout")
 	}
-	got = stdout.String()
+	got = stderr.String()
```
pmcatominey, line 1239: These changes were required after changing the timeout handling code in `template.go`, now that it returns an error the message is sent to stderr.

pmcatominey:
Updated to switch test from `SIGINFO` to `SIGHUP` as the former does not exist on linux/amd64.

flowchartsman:
You may want to consider holding off on this until `signal.NotifyContext` lands: https://github.com/golang/go/issues/37255

natefinch:
Sorry for the late review.

re: signal.NotifyContext .... I don't want mage to require a new version of Go, because I know the pain of being stuck on an older version.  I think we can do fine with the knobs we have.

This PR looks pretty good, but I think we're missing a synchronization effort... we should wait for the targets to clean up, and right now there's no mechanism for that.  So, like, if you put a defer in a target that takes a little time to run, it may not always have time to run before mage exits out.  I think we need a waitgroup and a timeout... like, first time we get sigint, cancel the context and start waiting for people to clean up.  Maybe wait up to 5 seconds.  If the user hits ctrl-c again, skip the timeout and just exit.

so something like this:

```
$ mage runlongthing
^C cancelling mage targets, waiting up to 5 seconds for cleanup...
^C exiting mage
$ 
```

pmcatominey:
@natefinch thanks for the review.

I've updated the code to apply a timeout after a SIGINT and tests to cover this, aswell as testing that a deferred function call within a target will be ran during during cleanup.

I'm not sure where a WaitGroup would be useful?

natefinch:
Thanks for pinging me on this, @faiq ... I'll have to re-review the code and work on the merge conflicts, but it seems like it should be able to be merged soon.  I'll try to poke at this some tonight.

natefinch (dismissed):
LGTM after a slight tweak.

Issue #323, opened 2020-12-07
target package: early exit and time comparisons


jasonmf:
Currently, the `target` package does some unnecessary filesystem inspection. It will recurse a directory or inspect a set of source files to collect the newest timestamp then compare that timestamp to a target to see if any of the source timestamps is newer than the target. The logic should be reworked to recurse/inspect, comparing against the target, and stop as soon as a newer source is found.

It would also be useful if the `Dir`, `Glob`, and `Path` functions supported supplying a timestamp as the target, rather than a file path. This would allow for comparison against arbitrary targets like container images using those functions. The existing three functions could then discover the target timestamp for comparison and just wrap functions that take a target timestamp.

PR inbound

AgentZombie:
Created https://github.com/magefile/mage/pull/324

natefinch:
Fixed! Thank you!

PR #324, opened 2020-12-07
target: rework modtime comparison logic #323

jasonmf:
Within the `target` package:

- Separate discovery of newest target timestamp from source inspection
- Allow source inspection functions to bail out as soon as they hit a timestamp newer than the target

```
~/p/mage/target$ go test -count=1 .
ok      github.com/magefile/mage/target 0.073s
```

### DIFF
--- a/target/newer.go
+++ b/target/newer.go
@@ -113,7 +113,7 @@ func NewestModTime(targets ...string) (time.Time, error) {
 				return err
 			}
 			mTime := info.ModTime()
-			if mTime.Before(t) {
+			if mTime.After(t) {
 				t = mTime
 			}
 			return nil
--- a/target/newer.go
+++ b/target/newer.go
@@ -1,0 +1,126 @@
+package target
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+var (
+	// errNewer is an ugly sentinel error to cause filepath.Walk to abort
+	// as soon as a newer file is encountered
+	errNewer = fmt.Errorf("newer item encountered")
+)
+
+// DirNewer reports whether any item in sources is newer than the target time.
+// Sources are searched recursively and searching stops as soon as any entry
+// is newer than the target.
+func DirNewer(target time.Time, sources ...string) (bool, error) {
+	walkFn := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.ModTime().After(target) {
+			return errNewer
+		}
+		return nil
+	}
+	for _, source := range sources {
+		source = os.ExpandEnv(source)
+		err := filepath.Walk(source, walkFn)
+		if err == nil {
+			continue
+		}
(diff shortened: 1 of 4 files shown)


DavidGamba:
I love this separate approach, the only other thing that I would add to this, which is the reason I don't use mage target is because in the logs I want to see a report of the file I found that was modified. Sometimes there are complex dependency trees and knowing what triggered what is helpful.

AgentZombie:
I started out trying to make it compatible with the upcoming `io/fs` interfaces to make it easy to write tests that didn't have to touch the filesystem. This would also allow user implementations that return `atime`, `ctime`, or other arbitrary times. Ultimately I saw there were already perfectly good unit tests so I got lazy. I might revisit this when `io/fs` lands and then users can inject whatever funky filesystem behavior they want.

target/target.go
```diff
@@ -77,100 +44,20 @@ func Glob(dst string, globs ...string) (bool, error) {
...
+	dst = os.ExpandEnv(dst)
+	stat, err := os.Stat(dst)
 	if os.IsNotExist(err) {
 		return true, nil
 	}
 	if err != nil {
 		return false, err
 	}
-	srcTime := stat.ModTime()
+	destTime := stat.ModTime()
```
natefinch, line 55: wow good catch, that was completely the wrong variable name

target/target.go
```diff
@@ -77,100 +44,20 @@ func Glob(dst string, globs ...string) (bool, error) {
 // doesn't exist, it always returns true and nil.  It's an error if any of the
```
natefinch, line 44: Can you edit the second sentence of the comment so it says

If a source or destination is a directory, this function returns true if a source has any file that has been modified more recently than the most recently modified file in dst.

I think that would make the behavior more clear.

target/newer.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		return nil
+	}
+	for _, source := range sources {
+		source = os.ExpandEnv(source)
+		err := filepath.Walk(source, walkFn)
+		if err == nil {
+			continue
+		}
+		if errors.Is(err, errNewer) {
```
natefinch, line 36: So I don't think filepath munges the error, and errors.Is isn't in older versions of Go, so I think I'd rather if this were just a plain old comparison.

```suggestion
		if err == errNewer {
```

natefinch (changes requested):
Mostly LGTM with a couple tiny tweaks

target/target.go
jasonmf: Updated and reflowed

target/newer.go
jasonmf: done

jasonmf:
Requested changes pushed

natefinch:
So, coming back to this, now that I better grok why there's an expanded API surface, I'm a little more ok with it. Letting people punch in their own modtime does make the code more flexible.

natefinch (approved):

PR #328, opened 2020-12-31
Add command line flag support

ryanfaerman:
This builds on the work that was recently merged, adding argument support. I've extended it a bit, adding support for optional flags in addition to positional arguments.

Given the following function in a magefile:
```go
func Say(animal string, loud mg.BoolFlag, msgs mg.StringSliceFlag) {
        fmt.Println("what does the", animal, "say?")
	if loud {
		fmt.Println("GET LOUD")
	}

	for _, m := range msgs {
		if loud {
			m = strings.ToUpper(m)
		}
		fmt.Println(m)
	}
}
```

It can now be executed as any of the following:

```
mage say fox --loud --msgs="bark bark" --msgs="woof woof" --msgs="yip yip"
mage say fox --msgs="bark bark" --msgs="woof woof" --msgs="yip yip"
mage say --msgs="bark bark" --msgs="woof woof" --msgs="yip yip" fox
mage say fox --msgs="bark bark"
```

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -23,15 +23,86 @@ import (
 	{{range .Imports}}{{.UniqueName}} "{{.Path}}"
 	{{end}}
 )
+func stripFlags(fs flag.FlagSet, args []string) []string {
+	if len(args) == 0 {
+		return args
+	}
+
+	commands := []string{}
+
+Loop:
+	for len(args) > 0 {
+		s := args[0]
+		args = args[1:]
+		switch {
+		case fs.Lookup(strings.Trim(s, "-")) != nil:
+			commands = append(commands, s)
+		case s == "--":
+			// "--" terminates the flags
+			break Loop
+		case strings.HasPrefix(s, "--") && !strings.Contains(s, "="):
+			// If '--flag arg' then
+			// delete arg from args.
+			fallthrough // (do the same as below)
+		case strings.HasPrefix(s, "-") && !strings.Contains(s, "=") && len(s) == 2:
+			// If '-f arg' then
+			// delete 'arg' from args or break the loop if len(args) <= 1.
+			if len(args) <= 1 {
+				break Loop
+			} else {
+				args = args[0:]
+				continue
+			}
+		case strings.HasPrefix(s, "-") && strings.Contains(s, "="):
+			parts := strings.SplitN(s, "=",2)
+			if fs.Lookup(strings.Trim(parts[0], "-")) != nil {
+			  commands = append(commands, s)
+			  continue
+			}
+		case s != "" && !strings.HasPrefix(s, "-"):
+			commands = append(commands, s)
+		}
+	}
+
+	return commands
+}
+
+func getFlags(args []string) map[string][]string {
+	flags := make(map[string][]string)
+	if len(args) == 0 {
+		return flags
+	}
+
+	for _, s := range args {
+		if s == "==" {
+			break
+		}
+
+		if !strings.HasPrefix(s, "-") {
+		  continue
+		}
+
+		parts := strings.SplitN(s, "=", 2)
+		key := strings.Trim(parts[0], "-")
+		if len(parts) == 2 {
+			flags[key] = append(flags[key], parts[1])
+		} else {
+			flags[key] = append(flags[key], "true")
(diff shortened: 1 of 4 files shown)


parse/parse.go
```diff
@@ -56,6 +56,34 @@ type Arg struct {
...
+	return strings.HasPrefix(a.Type, "mg") && strings.HasSuffix(a.Type, "Flag")
+}
+
+// FlagType returns the underlying type for the flag (string, bool, etc)
+func (a Arg) FlagType() string {
+	return strings.TrimSuffix(strings.TrimPrefix(a.Type, "mg."), "Flag")
+}
+
+// ZeroValue returns code needed to set the default value for a flag when it is
+// undefined. It needs to be a value that can be converted to the correct tyoe
```
gdey, line 70: [nit] Typo on the work 'type', on line 70

natefinch:
So the flags become optional parameters that you need to use `--` to specify?  That's interesting... but I'm not 100% sure the extra complexity in the code and the docs is worth it. Lemme think about it.

arschles:
@natefinch FWIW, I do think this is worthwhile. having named arguments is often more clear to users because they're more explicit. I've also had need for optional parameters to a target that I had to use environment variables for, and that's less discoverable for users.

sheldonhull:
I'll add a quick note to agree with @arschles. 
The _one_ thing that I was really surprised at and actually made things initially more complicated in adopting mage was lack of standard flag handling. I'd be fine with `param=1 mage do` or sticking with standard `mage do -filter "foo"` as well. I have tried to avoid parameters for any tasks where possible because it seems I can't make them optional right now (maybe I'm wrong there), and it's confusing to type parameters as if they are commands. 

If possible, I'd wish to support single dash flags since that's so common in Go, but double dash is also great.

mage/template.go
```diff
@@ -23,15 +23,86 @@ import (
...
 	{{end}}
 )
+func stripFlags(fs flag.FlagSet, args []string) []string {
+	if len(args) == 0 {
+		return args
+	}
+
+	commands := []string{}
+
+Loop:
```
sheldonhull, line 33: It's a `goto` in the wild, like finding a yeti 😆

parse/parse.go
sheldonhull, line 70:
```suggestion
// undefined. It needs to be a value that can be converted to the correct typo
```

giladreich:
@ryanfaerman, thank you so much for this.

@natefinch 👋 , can we please consider to add this? I've read through different threads (e.g. https://github.com/magefile/mage/issues/24) and to me it feels as this is the right approach, given that Golang [doesn't support overloading methods or optional arguments](https://golang.org/doc/faq#overloading).

At the time of writing I have to do this:
```go
func Test() error {
	pkgName := os.Getenv("PKG")
	filterRegex := os.Getenv("FILTER")
	if len(pkgName) != 0 && len(filterRegex) != 0 {
		return sh.Run("go", "test", "-v", "./"+pkgName, "-run", filterRegex)
	} else if len(pkgName) != 0 {
		return sh.Run("go", "test", "-v", "./"+pkgName)
	}
	return sh.Run("go", "test", "-v", "./...")
}
```

And invoke it like this:
```sh
PKG=util FILTER="Test*" mage test
```

Doesn't feel as intuitive or perhaps it's just me doing it wrong.

JeanChristopheMorinPerso:
Hello, just stopping by to say that I would be interested by this PR! The main advantage to me is the discoverability of target options/flags and better support for Windows (where `SOMETHING=asd mage ...` is not valid).

natefinch:
I definitely see the value in this, but I think I want to go a slightly different way.

I think we could dispense with mg.BoolFlag and just use pointers (*bool etc) instead. I think that's obvious enough to a go developer that *bool might be true, false, or unset.

c0nfleis:
What's the status on this? We build internal tooling using mage, and we have had to use a plethora of ENV vars in place of flags, which can be very convoluted at times :/

natefinch:
I'll spend some time on it this weekend, see if I can knock it out. I know it would be very useful.

natefinch:
I just shipped a version of this!

PR #345, opened 2021-05-06
Add support for running tasks on shutdown

a-palchikov:
This PR adds support for arbitrary tasks to run on shutdown. It is useful when the builder needs to clean up on exit.

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -20,6 +20,9 @@ import (
 	"strings"
 	"text/tabwriter"
 	"time"
+	
+	"github.com/magefile/mage/mg"
+
 	{{range .Imports}}{{.UniqueName}} "{{.Path}}"
 	{{end}}
 )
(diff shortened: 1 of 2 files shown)


natefinch:
Yeah, I think it makes sense. Maybe implemented like testing.Tcs Cleanup, so you can treat it like defer, but for the whole process run.

natefinch:
Yeah, so, I think it would be better to make a mg.Cleanup() method that stores a list of functions to call that are called in a defer in the generated main functions. 

The deinit is fine, but it means that deinit doesn't know what code has run. It would be better to have a place where you can shove code as you need it, just like defer.

jippi:
I like `mg.Cleanup` a lot better to be honest! :)

a-palchikov:
Agree on mg.Cleanup - I can take a stab at it in this PR.

jippi:
Sounds fantastic @a-palchikov - thank you! :)

PR #346, opened 2021-05-07
Detect deps that have already been invoked

### LINKED ISSUE #332: Improve error reporting on mg.Deps() usage errors
Given the following magefile:

```go
// +build mage

package main

import (
        "github.com/magefile/mage/mg"
        "github.com/magefile/mage/sh"
)

func Echo() error {
        return sh.RunV("echo", "hello")
}

func EchoDep() error {
        mg.Deps(mg.F(Echo()))
        return nil
}
```

```
$ mage echo
hello

$ mage echoDep
hello
Error: runtime error: invalid memory address or nil pointer dereference
```

carolynvs:
Detect when a dependency was accidentally already invoked

```go
mg.Deps(MyDep())
```

Normally this causes a nil pointer panic when we invoke the function

```
Error: runtime error: invalid memory address or nil pointer dereference
```

Now it will print the following as soon as its detected

```
Error: A dependency of the current target was defined improperly, with parenthesis. Dependencies should be defined as mg.Deps(MyDep), not mg.Deps(MyDep())
```

I have applied this change to checkfn so that it applies to serial and parallel deps.

Fixes #332


mg/deps.go
```diff
@@ -130,6 +130,9 @@ func runDeps(ctx context.Context, fns []Fn) {
 func checkFns(fns []interface{}) []Fn {
 	funcs := make([]Fn, len(fns))
 	for i, f := range fns {
+		if f == nil {
+			panic("A dependency of the current target was defined improperly, with parenthesis. Dependencies should be defined as mg.Deps(MyDep), not mg.Deps(MyDep())")
```
arschles, line 134: is panic the right thing here, or can we log and `os.Exit(1)` or similar?

arschles (approved):
LGTM. my only comment probably doesn't need to be addressed in this PR (at least IMO)

mg/deps.go
carolynvs: I used panic here because mage has a recover handler for panics that prints the error and exits with a non-zero code.

I just tried this branch again after the latest changes from main were merged and my patch doesn't seem to be working anymore though (it just returns the nil pointer instead now... not sure what changed to break the patch). So I'll have to take a look this weekend.

Were you seeing this panic not behave correctly that prompted you to add this comment? Just wondering if you are seeing different behavior with this patch than I am.

sheldonhull:
I ran into this today, thinking it was an issue with my logging. 
Once I changed like below it went away, so thanks for this tip. 

```
mg.Deps(
Js{}.Build(),
BuildAll(),
)
```

```
mg.Deps(
Js{}.Build,
BuildAll,
)
```

I'd like to see this merge get moved forward to support better error output.

carolynvs:
@natefinch Let me know if you'd prefer to see this fixed in another way. I'm happy to change it up if needed.

natefinch:
I think this is good but I think the error message needs to change a bit. It could be confusing if you pass in just a nil value of some sort.

I think something more like 
`fmt.Sprintf("mg.Deps expects a function, but got a nil value of type %T", f)`
would work in the more general case.

carolynvs:
Thanks for the feedback! 🙇‍♀️ 

After merging with master, I took another look and found a better place to prevent the nil-pointer error. It can occur with mg.F and mg.Deps/SerialDeps.

I incorporated your suggestion for the error message returned by mg.Deps. It doesn't assume that a nil must have come from an invoked function (though that's pretty common). Maybe they passed mg.Deps(3) instead and this will handle that better too.

```
non-function used as a target dependency: <nil>. The mg.Deps, mg.SerialDeps and mg.CtxDeps functions accept function names, such as mg.Deps(TargetA, TargetB)
```

When someone calls mg.F directly and passes in an invoked function, this is the error message:

```
non-function passed to mg.F: <nil>. The mg.F function accept function names, such as mg.F(TargetA, "arg1", "arg2")
```

natefinch (approved):
LGTM, thanks!

PR #347, opened 2021-05-17
feat: Limit parallel deps

ClaudiaJ:
Adds new flag `-p` to set maximum parallel deps.
Adds new environment `MAGEFILE_MAX_PARALLEL` set by the same flag.
Adds `mg.MaxParallel()` to parse the max parallel setting in targets.
Adds semaphore to `runDeps` configurable by max parallel setting to limit number of parallel deps.

Failsafe on new configured maximum parallel deps defaults to the maximum of either 1 or `runtime.NumCPU() - 1`

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -59,8 +59,10 @@ var mainfileTemplate = template.Must(template.New("").Funcs(map[string]interface
 }).Parse(mageMainfileTplString))
 var initOutput = template.Must(template.New("").Parse(mageTpl))
 
-const mainfile = "mage_output_file.go"
-const initFile = "magefile.go"
+const (
+	mainfile = "mage_output_file.go"
+	initFile = "magefile.go"
+)
 
 var debug = log.New(ioutil.Discard, "DEBUG: ", log.Ltime|log.Lmicroseconds)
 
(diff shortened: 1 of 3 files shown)


natefinch:
This unfortunately only limits the dependencies of a single function. transitive dependencies get their own count.  I think that would be surprising.  If I say p = 1, then I assume I'm only going to run the leaf-most dependency.  I think we can do this, but it'll have to be a different way.

arschles:
@natefinch do you think this is worth pursuing? I haven't had a use case for it personally, so I'd be interested in hearing if others do.

natefinch:
I've had a couple people ask for it, but it feels like a lot of work for an edge case, so I'm not sure.

mirogta:
We don't have a use case for it for any of our projects, where we could run multiple deps in parallel but they wouldn't give us much benefit. So we run only a couple in parallel - things that make difference, e.g. longish suites of tests split into groups via tags which can run independently. Even for the biggest projects we didn't have a need to run more than like 4 deps in parallel. The bottleneck there isn't parallelism, but downloading dependencies. Hope this helps.

Issue #351, opened 2021-05-27
Specify `GOARM`: is there a reason this is not supported?


qdm12:
You can specify `GOOS` and `GOARCH`, but not `GOARM` to build the binary?

I already forked the repository and ready to do a PR, but I wanted to check with you if there is an existing reason perhaps why this is not supported?

Thanks!

PR #352, opened 2021-05-27
Support for GOARM

qdm12:
Refers to question/issue #351 

It should all work nicely, although it adds that x/sys dependency to detect GOARM at runtime if it's not set, since it's not part of the `runtime` package. Maybe a better solution could be found. If we stick to that Golang solution, we may want to bump to Go 1.16 to have the latest x/sys dependency perhaps.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/magefile/mage
 
 go 1.12
+
+require golang.org/x/sys v0.0.0-20210525143221-35b2ab0089ea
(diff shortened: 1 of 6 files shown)


natefinch:
So, I am in favor of supporting this, but this would be the first time we have had an external import in mage.... and I really want to try to avoid them.  Do we *need* to set a default GOARM if one wasn't set from the command line?  I am happy to set GOARM if someone calls -goarm=6 ... but do things break if we don't set this manually?

qdm12:
To be honest, I don't exactly remember but from what I wrote.... 

I was trying to use mage on an emulated armv6 cpu and it would fail to build for armv6 I think. If you care to support building on armv6 natively, this PR might still have a role. You can check using Docker's `--platform=linux/arm/v6` to emulate this and use mage in that container.

Now passing a `-goarm` flag is needed for cross compilation. And cross compilation doesn't need this PR I think, since auto-detecting GOARM would not work if cross compiling from amd64 for example.

natefinch:
I think we autodetect so that people can set GOOS and GOARCH etc when running mage to build other code, and not screw up mage building its own binaries.  I guess we could just unset GOOS and GOARCH and GOARM and let the go tool detect it itself.

qdm12:
Yes that makes sense to me if that's the only reason, since the go tool detects it nicely already.

Maybe let's do that simplification first before finishing this PR for consistency (and to keep that zero dependency you have)?

PR #356, opened 2021-06-23
Collect both stdout and stderr in mage/sh

coryrc:
sh.BothOutput and sh.BothOutputWith

follows existing naming conventions.

### DIFF
--- a/sh/cmd.go
+++ b/sh/cmd.go
@@ -89,6 +89,19 @@ func OutputWith(env map[string]string, cmd string, args ...string) (string, erro
 	return strings.TrimSuffix(buf.String(), "\n"), err
 }
 
+// BothOutput runs the command and returns the text from stdout and stderr.
+func BothOutput(cmd string, args ...string) (string, string, error) {
+	return BothOutputWith(nil, cmd, args...)
+}
+
+// BothOutputWith is like RunWith, but returns what is written to stdout and stderr.
+func BothOutputWith(env map[string]string, cmd string, args ...string) (string, string, error) {
+	outBuf := &bytes.Buffer{}
+	errBuf := &bytes.Buffer{}
+	_, err := sh.Exec(env, outBuf, errBuf, cmd, args...)
+	return strings.TrimSuffix(outBuf.String(), "\n"), strings.TrimSuffix(errBuf.String(), "\n"), err
+}
+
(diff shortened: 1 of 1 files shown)


sh/cmd.go
```diff
@@ -89,6 +89,19 @@ func OutputWith(env map[string]string, cmd string, args ...string) (string, erro
 	return strings.TrimSuffix(buf.String(), "\n"), err
 }
 
+// BothOutput runs the command and returns the text from stdout and stderr.
+func BothOutput(cmd string, args ...string) (string, string, error) {
```
DavidGamba, line 93: `CombinedOutput` would make more sense as the name of this function.
https://golang.org/pkg/os/exec/#Cmd.CombinedOutput

coryrc:
@davidgamba (sorry, I can't find the comment in the UI, only e-mail!?)

CombinedOutput returns stdout and stderr in one stream, here we're returning them as separate streams. There isn't precedent in either os/exec or this package so I thought "BothOutput" but not wedded to it, of course.

DavidGamba:
@coryrc I sent that message before reading the code more in depth and that is why I deleted it afterwards, I realized it made no sense since the use case is different.

coryrc:
@DavidGamba Thought I was losing it!

Hey, since you're here, what do you think of *Output*V() versions which simultaneously stream to os.Stdout and os.Stderr?
Or are we just encoding all permutations of exec.Cmd in variable names?

DavidGamba:
Adding a *V version to `BothOutput` doesn't seem necessary since you are capturing the output, you could print it right after (though it won't be "live").

What is your use case here? I haven't found many cases where capturing stderr out of context within stdout is what I want and I would love to know how you plan to use this.

coryrc:
As part of CI tests I can popup errors to the user why it stopped. Stdout can have too much context and just Stderr is often enough to figure out what went wrong without having to page through a lot of output.

Without live streaming, you don't get any error message if the process is terminated due to timeout, which makes diagnosing hangs a real pain. Also an annoying user experience running locally, because it's hard to know if something is hung or slow.

jufemaiz:
@coryrc any further thoughts on the issues you've identified?

jufemaiz:
Aaaah fair enough.

Issue #288, opened 2020-02-06
Explicitly mention availability of brew installation in docs


neilotoole:
There's a `brew` formula available for `mage`: it should be explicitly mentioned in the repo `README.md` and on [magefile.org](https://magefile.org).

It really makes life easier to sell the usage of `mage` to skeptics if you can say to do this:

```
$ cd my/repo
$ brew install mage
$ mage install
```

Previously I'd been go-getting `mage` and using `bootstrap.go` etc... that was more effort than required (it's so simple with `brew`) and I saw pushback against the use of `mage` for that reason. I didn't know there was a `brew` formula. It's not in the docs, or at least, not obvious.

The website docs and `README.md` should be updated to reflect how easy it is to install `mage` via `brew`.

P.S. `mage` is wonderful. I was a minor skeptic initially, but I'm sold. Well done.

Related: #234

natefinch:
Ahh, cool, I honestly didn't know there was a brew formula for it :)  I should add that to how we release, and yeah, definitely add it to the docs.

PR #360, opened 2021-08-06
docs: add go install directions for Go 1.16+

sheldonhull:
- Provide `go install` command as option for those running Go 1.16+.

[docs](https://golang.org/ref/mod#go-install)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,13 @@ and Mage automatically uses them as Makefile-like runnable targets.
 Mage has no dependencies outside the Go standard library, and builds with Go 1.7
 and above (possibly even lower versions, but they're not regularly tested).
 
+
+**Using Go 1.16+**
+
+```
+go install github.com/magefile/mage@latest
+```
+
 **Using GOPATH**
 
 ```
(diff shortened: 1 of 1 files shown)


sheldonhull:
@alock goreleaser is just a tool for building and pushing various distribution methods. It provides a nice way to automate different distribution methods like hombrew, curl single line install scripts, docker images, and stuff like that. Not really solving the same issue I was trying to provide by just adding the native go install technique.

@natefinch the version metadata could be set to the git commit sha when not provided by a semver version. Would that be acceptable? Go install just provides a really low overhead cross platform way to install this so figured would be worth trying.

arschles:
>ideally, people wouldn't install that way, because it makes it harder to debug problems without that build info.

+1. I think mage would be better served by having homebrew/apt/$PACKAGE_MANAGER packages and including install instructions for those, instead of asking people to install via the `go` tool

sheldonhull:
We can look at that. Makes sense to have a curl install option. However, considering go install solves cross platform setup for Go devs without any need to worry about handling packaging/release of binaries, I thought it was worth considering. I bootstrap all my dev environments with a range of `go install` cli tools. 

I guess this can be closed out.
New work for Scoop for windows, brew for mac, and curl for linux perhaps in the future. :thumbsup:

arschles:
@sheldonhull fair enough. at some point in the future, `brew install mage` would be 🤯

natefinch:
brew install is probably good to have

mirogta:
> brew install is probably good to have

@natefinch it already exists - see [here](https://github.com/magefile/mage/issues/288) and [here](https://formulae.brew.sh/formula/mage) - but I can't see who owns it and how does it get updated

PR #361, opened 2021-08-06
chore(project): 🔧 add editor config and line normalization

sheldonhull:
Feel free to reject this as not a priority change, just something that _might_ help as I noticed a recent docs update flagged a line ending change unnecessarily when I updated README.md.

Recently discovered contributions can have some conflicting behavior depending on macOS, Linux, or Windows as the source system.

With gitattributes, the conflict of line endings can be solved without worrying about the user's personal gitconfig settings by including the line normalization in the actual gitattributes file. 

> The advantage of this is that your end of line configuration now travels with your repository and you don't need to worry about whether or not collaborators have the proper global settings.

The editor config can also help eliminate conflicts by ensuing personal editor preferences don't conflict between other contributors. 

[github docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings)


sheldonhull:
need to revert the readme update as forgot to reset this as a separate PR.

mirogta (approved):
LGTM - @natefinch I've tested this on MacOS and Win, if you're happy to get it merged.

natefinch:
Yeah I'll look at this again tonight. I have some reservations about random config files for a million things.... But I appreciate the attempt to keep things consistent.

.editorconfig
```diff
@@ -0,0 +1,30 @@
...
+
+# Makefiles always use tabs for indentation
+[{Makefile, makefile, GNUmakefile, Makefile.*}]
+indent_style = tab
+
+[*.md]
+trim_trailing_whitespace = true
+
+[*.sh]
+indent_style = space
```
thaney071, line 29: Suggestion: indent_style is already `space` for the root setting. setting it again for `sh` isn't needed.

.editorconfig
```diff
@@ -0,0 +1,30 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
```
thaney071, line 5: I know gofmt will fix this, but go uses tabs not spaces. Might be good to set that in here. I am not sure if go cares what indent size is used for tabs. I have mostly seen it at 4 and gofmt might just do that anyway.

Consider adding a section for go, or adding to an existing one where makes sense.

sheldonhull:
💀

Issue #371, opened 2021-09-24
Target is launched which should not be launched


whiterm:
To reproduce this bug, it was necessary to run `go run mage.go test4` several times, because this error is floating
```
// +build mage

package main

import (
	"fmt"

	"github.com/magefile/mage/mg"
)

func Test1() error {
	fmt.Println("execute test1")
	return fmt.Errorf("error test1")
}

func Test2() error {
	mg.Deps(Test1)
	fmt.Println("execute test2")
	return nil
}

func Test3() error {
	mg.Deps(Test2)
	fmt.Println("execute test3")
	return nil
}
func Test4() error {
	mg.Deps(Test1, Test2, Test3)
	fmt.Println("execute test4")
	return nil
}
```
Output:
```
execute test1
execute test3
Error: error test1
error test1
exit status 1
```
`test3` target is not expected to run

natefinch:
Ooh crap, I think I see the problem. If test1 runs and errors out, it's still marked as "run", so test3 sees that it doesn't need to run test1 and assumes it had succeeded.  Instead we need to mark test1 as a failure so that if anyone else tries to run it, they fail out.

whiterm:
I think you are right. I just changed the test code a little bit. And now the bug can be reproduced at every run
```
// +build mage

package main

import (
	"fmt"
	"time"

	"github.com/magefile/mage/mg"
)

func Test1() error {
	fmt.Println("execute test1")
	return fmt.Errorf("error test1")
}

func Test2() error {
	mg.Deps(Test1)
	fmt.Println("execute test2")
	return nil
}

func Test3() error {
	time.Sleep(time.Second)
	mg.Deps(Test2)
	fmt.Println("execute test3")
	return nil
}
func Test4() error {
	mg.Deps(Test1, Test2, Test3)
	fmt.Println("execute test4")
	return nil
}
```

PR #373, opened 2021-09-27
Fix racy dependencies by replaying the panic

a-palchikov:
Persist the panic that a dependency had failed with and replay it each time (*onceFun).run is invoked.
This fixes the behavior when a dependency is tested concurrently. Using this exerpt from the issue example, if the Test2 is invoked concurrently with Test3->Test2, and the first Test2 gets to run the target (via [sync.Once.Do](https://pkg.go.dev/sync#Once.Do)), it properly propagates the failure from Test1, but the 2nd Test2 (executed via Test3) will not get the same handling as no panic happens and it simply returns with `nil` as it does not itself return an error:
```go
Test4() {
  mg.Deps(Test1, Test2, Test3)
}
```
I took the liberty of restructuring `runDeps` to be more idiomatic by using channels but let me know if that's beyond the scope.

Fixes https://github.com/magefile/mage/issues/371.


arschles (dismissed):

jippi:
I've been hit with this issue a couple of times, @natefinch do you think it need more work before being merged in?

natefinch:
Let me look at this tonight. I'll make time for it. Sorry for the delay

Issue #209, opened 2018-12-13
Override targets from imported Magefile


anonfunc:
An example would be importing a package with a known Build and Deploy, and overriding the Deploy to add additional logic.

Right now, I do this by removing the "// mage: import" comment and adding a bunch of new functions like:
```
// Cleans known output paths.
func Clean() error {
	mg.Deps(imported.Clean)
	return nil
}
```

Ideally, my Clean would take priority over an imported Clean.

I haven't dug too much into the internals, my apologies if this is a crazy request.

natefinch:
That's not a crazy request, but I'll need to think about it. The problem with overriding is that then it can get confusing which target you're running.

What might work better is to give the function in the imported package a signature that won't be treated as a target, and then your Clean target can just call it like normal code.    ...  what might be useful in that case is a way to mark a function as explicitly not something you want treated as a target so you can give it a signature that still looks like a mage target.   Like

```
// mage:noexport
func Clean() error {

}
```

That assumes that it's ok to make everyone who imports the package write their own Clean target, which could then call this Clean function in the imported library.  If you want some people to get it imported and some not, then things get tricky, and overriding would work better.... but like I said, I'd need to think on that.

The other option, of course, is not to call it Clean, call it CleanFiles or CleanBase or something, and then write a Clean target in the importing magefile.   That'll give you two targets when you run `mage -l` but that might not be the end of the world.  But if you need to override a lot of imports, that gets annoying, too.

Sorry for the stream of consciousness replies...  I don't know exactly your use case, so I'm trying to get as many ideas out there as possible, in case one of them can be useful to you.

anonfunc:
Here's my use case, more explicitly:
I've got a variety of repos which are *mostly* Go projects with a standard layout and the same deployment command.  There are a few exceptions: one repo might have several related deployments, one repo might want to run something after the deployment, etc.

Previously, we were copying a Makefile around and editing it for the exceptional projects, but I decided that Mage looks perfect for this:  put the common build logic in a package, import it into each repo via the `mage:import` comments, everything builds the same way, build maintainers are happy.   (Mage also looks great for a bunch of other reasons, but this was a big reason.)

This leads to finding the "right" solution for the exceptions.  So far, we've considered (or implemented) the following:

- More granular target packages, so we can say "import build and test, but not deploy"
- Importing the target packages and delegating (snippet above)
[...]

PR #380, opened 2021-10-21
feat: add ability to skip importing target

viktorvoltaire:
Potential solution for some of the problems mentioned in https://github.com/magefile/mage/issues/209

With this PR we can have functions in packages that we import, that are more support functions but not necessarily mage targets. `skip` might not be clear enough, we could also go for something like `noExport` or `excludeTarget`..

- [x] Add test
- [ ] Add documentation

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -18,6 +18,7 @@ import (
 )
 
 const importTag = "mage:import"
+const skipTarget = "mage:skip"
 
 var debug = log.New(ioutil.Discard, "DEBUG: ", log.Ltime|log.Lmicroseconds)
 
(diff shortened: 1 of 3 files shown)


natefinch:
I don't like this, to be honest. It complicates things. If you want exported helpers, they can be in a different package. I really don't want a proliferation of special comments. The nice thing about mage is that you *don't* need a lot of specially comments. mage:import is literally the only one right now.

mirogta:
@viktorvoltaire would you be happy to close this PR without merging then? I agree with @natefinch that splitting helper functions into a different package seems trivial and cleaner.

viktorvoltaire:
> @viktorvoltaire would you be happy to close this PR without merging then? I agree with @natefinch that splitting helper functions into a different package seems trivial and cleaner.

Sure, we can always re-open it

PR #383, opened 2021-11-01
Add asdf installation instructions to docs

mathew-fleisch:
The [asdf version manager](https://asdf-vm.com/) is a tool for installing release binaries from Github. With asdf installed, the [asdf plugin for mage](https://github.com/mathew-fleisch/asdf-mage) can be used to install any released version of mage.

Please let me know if there is anything else I can do for this PR.

PS: Also willing to transfer ownership of the plugin if that is desirable to the community

### DIFF
--- a/site/content/index.md
+++ b/site/content/index.md
@@ -42,6 +42,27 @@ The mage binary will be created in your $GOPATH/bin directory.
 You may also install a binary release from our
 [releases](https://github.com/magefile/mage/releases) page. 
 
+**Using asdf**
+
+The [asdf version manager](https://asdf-vm.com/) is a tool for installing release binaries from Github. With asdf installed, the [asdf plugin for mage](https://github.com/mathew-fleisch/asdf-mage) can be used to install any released version of mage.
+
+```shell
+# Install asdf plugin for mage
+asdf plugin add mage
+
+# Show all installable versions
+asdf list-all mage
+
+# Install specific version
+asdf install mage latest
+
+# Set a version globally (on your ~/.tool-versions file)
+asdf global mage latest
+
+# Now mage commands are available
+mage --version
+```
+
 ## Example Magefile
 
 ```go
--- a/site/content/index.md
+++ b/site/content/index.md
@@ -62,20 +62,9 @@ See [scoop](https://scoop.sh/).
 The [asdf version manager](https://asdf-vm.com/) is a tool for installing release binaries from Github. With asdf installed, the [asdf plugin for mage](https://github.com/mathew-fleisch/asdf-mage) can be used to install any released version of mage.
 
 ```shell
-# Install asdf plugin for mage
 asdf plugin add mage
-
-# Show all installable versions
-asdf list-all mage
-
-# Install specific version
 asdf install mage latest
-
-# Set a version globally (on your ~/.tool-versions file)
 asdf global mage latest
-
-# Now mage commands are available
-mage --version
 ```
 
 ## Example Magefile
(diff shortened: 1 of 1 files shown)


site/content/index.md
```diff
@@ -57,6 +57,27 @@ See [mage homebrew formula](https://formulae.brew.sh/formula/mage).
 
 See [scoop](https://scoop.sh/).
 
+**Using asdf**
```
natefinch, line 60:
```suggestion
### Using asdf
```

site/content/index.md
```diff
@@ -57,6 +57,27 @@ See [mage homebrew formula](https://formulae.brew.sh/formula/mage).
...
+
+# Install specific version
+asdf install mage latest
+
+# Set a version globally (on your ~/.tool-versions file)
+asdf global mage latest
+
+# Now mage commands are available
+mage --version
+```
```
natefinch, line 79: since this is a specific installer that not everyone will use, let's try to keep the install sections short.  Can you make this a lot shorter, following the examples of the other installers above?  Like a link to the docs, and then just the 2-3 lines you need to run to install?

natefinch (changes requested):
Thanks for the contribution! Can you shorten it up a bit? :)

site/content/index.md
mathew-fleisch: done

site/content/index.md
mathew-fleisch: how's that?

natefinch (approved):
LGTM!

site/content/index.md
natefinch: looks good!

sheldonhull:
@natefinch can we get this transferred to the magefile org? This is really useful for projects and would be good to know the regular test suite runs. The test case is already documented too.

mathew-fleisch:
I'm fine with that. Other than copying the repo to the mage org, we'd also need to update the [asdf-plugins](https://github.com/asdf-vm/asdf-plugins/) repo to point at the new location.

natefinch:
I'm not entirely sure that asdf is even really needed for mage. It says it's intended to manage different versions of tools, but mage is 100% backwards compatible, so you don't need to manage versions, you can always just get latest and it'll work with any mage script ever written.

mathew-fleisch:
it's probably not "needed" 😏 but the reason I love to manage tools with asdf is it provides a standard way of installing binaries from a CI pov, that theoretically works on multiple system architectures. For instance, this tool is a github-action linter written in rust and I install the last stable version with asdf on commits to any PR to ensure changes to any other action conforms to standards.
https://github.com/mathew-fleisch/action-validator/blob/test-self-hosted/.github/workflows/self-hosted-example.yml#L23-L33

I don't personally regularly use mage, and I cannot remember why I originally made this plugin 🤔 , but someone might want to do the same and I'll transfer ownership if desired 🤷 

Also happy to continue "maintaining" this plugin, but there haven't been any PRs yet.

natefinch:
That's funny that you're not even using Mage. Sheldon, do you want to take over, since you actually do use it? I know the burden of maintaining stuff that you don't even use.

sheldonhull:
@natefinch be more than happy to help where I can. I've been meaning to explore creating an asdf plugin but hadn't yet. I actually do most of my installs in mage tasks now 😆 and avoid bash scripting where I can. I'll have to getfamiliarr with the asdf one.

sheldonhull:
> I'm not entirely sure that asdf is even really needed for mage. It says it's intended to manage different versions of tools, but mage is 100% backwards compatible, so you don't need to manage versions, you can always just get latest and it'll work with any mage script ever written.

To be fair I prefer `go install` for mage, as it's ok to not have the commit info for me mostly.
That said, the point of asdf is that common tools are just easy to immediately access. 

The versioning is a perk, but running on latest is also totally fine.
It just makes it quick and simple to setup a container or project with all the tooling.

PR #391, opened 2021-12-10
ci: migrate from travis to github action

### LINKED ISSUE #378: Update CI to use stable 1.16 go
See https://github.com/magefile/mage/blob/dd944245e253451de4f2f512ec596a5c6f05377b/.travis.yml#L12

Makes me think only `1.16beta1` was available at the time the Travis CI config was updated.

Any objections to bump it to `1.16.1` ?

chenrui333:
- Add github action workflow (see build results in here, https://github.com/chenrui333/mage/actions)
- Remove travis as travis does not build the project anymore

---

closes #378


mirogta (approved):
LGTM

@natefinch - this drops go 1.9 and 1.10 from the tests. I know how strongly you feel about mage's backwards combability - are you also happy with removing those?

natefinch:
Is there any reason we can't support 1.9 and 1.10 as well?

chenrui333:
> Is there any reason we can't support 1.9 and 1.10 as well?

Go modules was introduced in go1.11, that is the main reason why it does not work for them.

natefinch:
Oh yeah, lol, yeah, that seems fine.

natefinch (approved):
LGTM

thaney071:
They should still be backwards compatible. The go tools changed behaviors at go 1.12 to be [module aware](https://go.dev/doc/go1.12#modules). 

> The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.
>
> This changed use of the go directive means that if you use Go 1.12 to build a module, thus recording go 1.12 in the go.mod file, you will get an error when attempting to build the same module with Go 1.11 through Go 1.11.3. Go 1.11.4 or later will work fine, as will releases older than Go 1.11. If you must use Go 1.11 through 1.11.3, you can avoid the problem by setting the language version to 1.11, using the Go 1.12 go tool, via go mod edit -go=1.11.

The CI for 1.11.x probably passes because it is defaulting to 1.11.4 which is supported by the repos go.mod version directive.

For builds in 1.10.x and 1.9.x, they need to be setup to be in the GOPATH for them to work. I see that you set the GOPATH variable for testing, but the code itself needs to checked out to the GOPATH for the to tooling to work, and it must be in the correct layout for the imports to work.  

I tried debugging a bit more, but am not as familiar with github actions. As far as I can tell the checkout is not following the expected layout for go version 1.10 and 1.9 to work. see: https://github.com/thaney071/mage/actions

Sorry, long post to say the builds should be able to still work using 1.10 and 1.9, but the checkout step needs to be configured so the code is checked out to the GOPATH as: `$GOPATH/src/github.com/mage/mage`

chenrui333:
> Sorry, long post to say the builds should be able to still work using 1.10 and 1.9, but the checkout step needs to be configured so the code is checked out to the GOPATH as: `$GOPATH/src/github.com/mage/mage`

Yeah, if we still want to make the builds successful with 1.9 or 1.10, I agree that we can configure GOPATH.

But soon or later, we will need to form a plan to remove go1.9 and go1.10 support.

chenrui333:
> This isn't getting picked up in the "checks" section.... I always forget if I need to do something special to turn these things on, or is it just that this is an outside contributor, so it doesn't actually run on the PR?

Since github action is not on the base branch yet, it did not get run. After the merge it would run on the PRs. (that is why I posted the action runs on my fork)

mirogta:
> But soon or later, we will need to form a plan to remove go1.9 and go1.10 support.

@natefinch In relation to this, have you thought about adding some anonymous telemetry collection when `mage` is executed, which could be of course turned off, but perhaps on by default?

OS version, go version, build duration, which features of mage are used? How many are using mage with go 1.9/1.10, or mage flags (`-w` etc.), or mage functions (`sh.Run` etc.), and more… To get a better picture than what we get just from people telling us here or in a chat. Something like https://yarnpkg.com/advanced/telemetry

thaney071:
If @natefinch and @mirogta are okay with not testing on go 1.9 and 1.10, this can be merged.  I can keep looking at the action setting or update the docs/web template for the tested versions.

The website has 1.7 as the lowest tested as of today.

natefinch:
I think testing on 1.11+ should be fine. That's fall 2018 and was when modules were first added as opt-in.

PR #398, opened 2021-12-22
feat: allow mage:import alias to be defined for multiple imports

viktorvoltaire:
When testing this locally, I haven't been able to trigger any collisions that isnt already flagged by the existing sanity checks. So far everything has just worked as expected, all existing tests are passing.  

Documentation wise, I couldnt find anything that actually specifies that the an import alias needs to be unique, so I dont feel we need to add something there

### DIFF
--- a/mage/import_test.go
+++ b/mage/import_test.go
@@ -312,3 +312,57 @@ Error parsing magefiles: error running "go list -f {{.Dir}}||{{.Name}} github.co
 		t.Fatalf("expected:\n%s\n\ngot:\n%s", expected, actualShortened)
 	}
 }
+
+func TestMageImportsSameNamespaceUniqueTargets(t *testing.T) {
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	inv := Invocation{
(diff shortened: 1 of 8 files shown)


viktorvoltaire:
@natefinch given our discussion on slack, should i proceed with writing a test where we expect error if target exist in both packages?

mirogta:
@viktorvoltaire wow, nice one, why hadn't I think about this one? :-)

Looks good.

Documentation wise: Now I wonder if you have a specific real use case where this would help you and it might me beneficial for others to explain this use case and actually mention when/how this can be used?

viktorvoltaire:
> Documentation wise: Now I wonder if you have a specific real use case where this would help you and it might me beneficial for others to explain this use case and actually mention when/how this can be used?

We use the namespaces to group commands, to make it easier to identify. For us it makes sense that imports from different packages should be able to be grouped together. I actually thought that this was possible before the issue showed itself for me :) Hence why I think the documentation is good as it is

viktorvoltaire:
@natefinch what do you think?

natefinch (approved):
LGTM, thanks for this!

PR #399, opened 2021-12-23
chore: wrap log.Println in cmd with verbose check

viktorvoltaire:
~~This was introduced in https://github.com/magefile/mage/pull/306 it feels like it was left there by mistake. Its causing a lot of unwanted log prints~~

Edit: This is triggered from doing `sh.Output/Run etc...` in a global variable.

### DIFF
--- a/sh/cmd.go
+++ b/sh/cmd.go
@@ -137,7 +137,10 @@ func run(env map[string]string, stdout, stderr io.Writer, cmd string, args ...st
 	for i := range args {
 		quoted = append(quoted, fmt.Sprintf("%q", args[i]));
 	}
-	log.Println("exec:", cmd,  strings.Join(quoted, " "))
+	// To protect against logging from doing exec in global variables
+	if mg.Verbose() {
+		log.Println("exec:", cmd, strings.Join(quoted, " "))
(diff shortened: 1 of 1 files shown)


viktorvoltaire:
@mirogta @natefinch

mirogta:
Looking at #306, it's a 2yr old issue.

So I'm actually looking at the main branch: https://github.com/magefile/mage/blob/master/sh/cmd.go#L140 which does have the `log` there.

I understand that some people might not like this extra logging, but it doesn't look like it was introduced in #306 and removing it might be seen as a breaking change. We try to prevent any breaking changes.

@natefinch could we wrap this line with a Verbose check, as an improvement/compromise?

viktorvoltaire:
@mirogta you are correct, I was to hot on the git blame. Ive been debugging this a bit with @natefinch and found out that the unwanted behavior comes from a global variable

viktorvoltaire:
@mirogta @natefinch I tested this locally and with the conditional I dont get any more output from doing exec in globals

natefinch (approved):
This looks like a good, clean fix. Thanks!

PR #402, opened 2022-01-14
Add variadic support to mg.F

### LINKED ISSUE #401: Enhancement: Variadic support in mg.F
**Describe the feature**
I would like `mg.F` to support variadic functions. 

**What problem does this feature address?**
The main use case would be to pass sh.Run to it, as such: `mg.F(sh.Run, "go", "test", "./...")`
I think this should be possible with some work. `mg.F(sh.Run, "go", []string{"test", "./..."})` would also be ok, or another function instead of `mg.F`.

Right now there's a lot of single-line functions that need to be added that are then passed to mg.Deps. As pointed out in #400, it's possible to use inline function literals for this, but those are also quite verbose. Also they might run the same command multiple times, as the arguments and actual function are not part of the deps check for those literals.

Having sh.Run directly in mg.Deps improve readability and simplifies the magefile writing.

**Additional context**
This occurred to me during the discussion in #400 where I had a different syntax for the same thing.

perj:
Allows to pass sh.Run to mg.F as such:

	mg.Deps(
		mg.F(sh.Run, "go", "test", "./..."),
	)

This improves the magefile by removing some of the one-liner functions
that you might end up with that are only used through mg.Deps.

Resolves #401.

### DIFF
--- a/mg/fn.go
+++ b/mg/fn.go
@@ -111,8 +111,8 @@ func checkF(target interface{}, args []interface{}) (hasContext, isNamespace boo
 		return false, false, fmt.Errorf("target's return value is not an error")
 	}
 
-	// more inputs than slots is always an error
-	if len(args) > t.NumIn() {
+	// more inputs than slots is an error if not variadic
+	if len(args) > t.NumIn() && !t.IsVariadic() {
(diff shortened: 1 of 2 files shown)


perj:
Hi @natefinch. Did you have any chance to look at this yet?

mirogta (approved):
LGTM

PR #407, opened 2022-02-24
bring docs for use of directives up to date

### LINKED ISSUE #406: doc: consider altering // mage:import documentation to remove leading space
The "mage:import" directive comment looks very much like it falls in the family of [directive comments](https://pkg.go.dev/cmd/compile#hdr-Compiler_Directives) used by the compiler and other tools, for example linters (see golang/go#43776).

However it specifies in the documentation that a leading space should be included. The code specifically [does not require this](https://github.com/magefile/mage/blob/526bf4775130d567de6b18fea8f3791d76c8be2a/parse/parse.go#L463-L465), so I would like to suggest that the documentation be changed to exclude the space.

This would have no backward incompatibility issues due to the current acceptance of the no-space variant, but over time would allow the Go community to converge on a consensus approach to using directive comments.

efd6:
Please take a look.

It might be worth running gofmt over the codebase to bring the uses of build tag directives up to date as well. I'm happy to do that in this PR if you would like.

Fixes #406.


natefinch (approved):
LGTM, thanks!

efd6:
Is there any more to do here?

natefinch (approved):
LGTM, thanks!

PR #411, opened 2022-03-17
Update site to build with newer hugo

perrito666:
I updated the site to build with latest (as of today) hugo.

Please check this looks as you expect


natefinch (dismissed):
LGTM. Thanks for all the help!

natefinch (approved):

PR #417, opened 2022-03-25
add version info automatically if you build with go 1.18+

natefinch:
The only reason we have an install script is to set build data. As of 1.18, we don't need that anymore. So, now the logic will be split, if you have 1.18, you can just do `go install github.com/magefile/mage@latest` and it'll Just Work™.

If you have go 1.17.x or earlier, you'll still need to run `go run install.go` to have the local mage binary report version info.

One slight suboptimalness (that's a word, I swear) of the 1.18 data is that it doesn't include tag info, only commit info. That's not very user-friendly, so I wrote code to get tag info from github... which means `mage --version` has to go out to the internet if it was build with 1.18. That stinks, but I think it's worth it if it means `go install` mostly just works.

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -66,13 +66,6 @@ const (
 
 var debug = log.New(ioutil.Discard, "DEBUG: ", log.Ltime|log.Lmicroseconds)
 
-// set by ldflags when you "mage build"
-var (
-	commitHash = "<not set>"
-	timestamp  = "<not set>"
-	gitTag     = "<not set>"
-)
-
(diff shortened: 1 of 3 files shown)


natefinch:
Evidently the version is in `BuildInfo.Main.Version` if you've used go install, but won't be there if you just build with go build from a locally cloned repo, so probably can just check that and if it exists, use it, otherwise do the network dance above.

natefinch:
This is old and I am already using that build info :)

Issue #286, opened 2020-01-24
rename imports in generated mainfile


natefinch:
The generated mage_output_file.go uses imports like "log" which can easily conflict with global variables in a user's magefile, so we should name them something unlikely to collide with people's own variables.

PR #421, opened 2022-04-14
feat: rename templated imports to avoid collisions

ladydascalie:
Addresses #286

This PR renames most imports in the template to minimise the risk of collisions.
Some imports (like `context`) may need to remain pure, for generated `wrapFuncs` to function properly (unless this also merits changing).

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -9,16 +9,16 @@ package main
 
 import (
 	"context"
-	"flag"
-	"fmt"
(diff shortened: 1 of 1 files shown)


natefinch:
This looks really good. I've been meaning to do this for a long time. Thanks for the work. 

I think it's fine without changing context for now.

natefinch (approved):
LGTM, Thanks!

PR #435, opened 2022-07-28
Using sh.Run instead of exec.Command in the magefile default template

### LINKED ISSUE #408: magefile generated init should use best practices
Not sure if it's literally a bug but it seems wird to me that the magefile you generate with `init` does not use the built in libraries like `sh` but instead they use standard library functions, e.g.

```
	cmd := exec.Command("go", "build", "-o", "MyApp", ".")
```

I could also make a PR for that but I wanted to check first whether it is intentional.

jespino:
Before this PR the template created by the command `mage -init` was using
`exec.Command` instead of `sh.Run`. Now this is more "natural" for mage users.

Fixes #408

### DIFF
--- a/mage/magefile_tmpl.go
+++ b/mage/magefile_tmpl.go
@@ -8,9 +8,9 @@ package main
 import (
 	"fmt"
 	"os"
-	"os/exec"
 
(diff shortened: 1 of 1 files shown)


mloskot:
Why this has been closed?
Isn't it still relevant to address the #406?

PR #436, opened 2022-07-28
Showing panic trace on debug mode

### LINKED ISSUE #333: Improve panic reporting
From a UX perspective, "Error: runtime error: invalid memory address or nil pointer dereference" is not particularly helpful if I've made a mistake in my magefile that causes a panic, since there's zero context to localize it. Though I appreciate the desire to keep error message parsimonious in a build tool, it would be super helpful if I could get a full stacktrace when running with `-debug` or at least the relevant location of the panic without `-debug`

jespino:
It enables the hability to show the whole trace for panic errors inside your
magefile when you are in debug mode. If you are not in debug mode, the behavior
keeps the same.

Fixes #333

### DIFF
--- a/mage/template.go
+++ b/mage/template.go
@@ -9,6 +9,8 @@ package main
 
 import (
 	"context"
+	_runtime_debug "runtime/debug"
+	_errors "errors"
 	_flag "flag"
 	_fmt "fmt"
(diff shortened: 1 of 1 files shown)


flowchartsman:
Is this good to merge?

PR #437, opened 2022-08-05
Proposal for binary cache cleanup

### LINKED ISSUE #319: Mage Cache Size Management
Someone at work was complaining that their .magefile directory was getting bloated with old binaries... which is a completely valid complaint. We should add autocleanup to delete old binaries in the magefile directory.

Something like

MAGEFILE_MAX_CACHE_AGE  - number of days. Default... 10?
Binaries in the mage cachedir that are older than this would get deleted whenever mage runs.

MAGEFILE_AUTOCLEANUP - default true, if true run autocleanup whenever mage runes (see max cache age)

MAGEFILE_MAX_CACHE_SIZE  - number of bytes. Default... 200?
If the mage cache size is larger than this, delete the oldest binaries in the directory until it goes under this number.

MAGE_NOCACHE - default false. if true, never cache binaries (i.e. delete them after running).  This would mean you always rebuild the binaries, which can be slower.

jespino:
This PR is a proposal to implement the cache auto-cleanup functionality using the environment variables suggested.

I have some doubts about how to define things there. I think the variables have 2 different granularity. For example, the size in bytes, makes no sense to me, because the go binaries are going to be around 2MB in the best case, so... would make more sense to use MB instead of bytes to measure that value. The age in days, sounds ok, not a big deal, but maybe we want more granularity there (or not).

Also, I did a "touch" of the binary files in every execution to use that information as part of the cleanup process, that way I can store in the filesystem if I need to clean up that or not.

Fixes #319

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
@@ -30,6 +31,9 @@ import (
 // changed. This can be used when we change how we parse files, or otherwise
 // change the inputs to the compiling process.
 const magicRebuildKey = "v0.3"
+
+const defaultMaxCacheSize = 200000000
+const defaultMaxCacheAge = 10
 
 // (Aaaa)(Bbbb) -> aaaaBbbb
 var firstWordRx = regexp.MustCompile(`^([[:upper:]][^[:upper:]]+)([[:upper:]].*)$`)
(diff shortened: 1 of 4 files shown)


mage/main.go
```diff
@@ -456,9 +460,64 @@ func Invoke(inv Invocation) int {
...
+		if f.IsDir() {
+			continue
+		}
+		totalSize += f.Size()
+		if maxCacheAge > 0 && timeLimit.After(f.ModTime()) {
+			os.Remove(filepath.Join(cacheDir, f.Name()))
+		}
+		if maxCacheSize > 0 && totalSize > maxCacheSize {
+			os.Remove(filepath.Join(cacheDir, f.Name()))
+		}
```
mgabeler-lee-6rs, line 516: The computed size of the cache shouldn't include files that are deleted -- it'd cause excess files to be removed.

```suggestion
		if maxCacheAge > 0 && timeLimit.After(f.ModTime()) {
			os.Remove(filepath.Join(cacheDir, f.Name()))
		}
		if maxCacheSize > 0 && totalSize + f.Size() > maxCacheSize {
			os.Remove(filepath.Join(cacheDir, f.Name()))
		}
		totalSize += f.Size()
```

jespino: Very good point, this is failing when you combine the 2 policies at the same time. Thanks! I'll add a test for that too.

jespino: Actually, I realized that is not the case, there are ordered by modification date, so... if you already remove one file because the modification date, you are going to remove every other file after that, so... doesn't matter if the size is affected or not. On the other hand, if the size is the one triggering the remove, you already surpassed the size, so the age is not relevant anymore... I think the code works well, anyway I going to do it as an `else if` to avoid duplicated `os.Reomve`.

mgabeler-lee-6rs: re the ordering: Consider a size limit of 10 and a sequence of cached files of sizes 2,9,1,1,1 (after timestamp ordering)

Your algorithm will delete everything except the first one. But in reality only the "9" needs to be deleted.

In general, anything that is deleted by definition does not contribute to the total size :)

jespino: Interesting point. I would expect that size of files are going to be pretty homogeneous, but its is true that you can have dependencies and in consequence generate pretty big binaries. Let me try to create a test case for that. I don't think the current policy is necessarily wrong (We can assume that everything that is after certain size, gets removed, and always by date), but I agree that can be better in that case.

Issue #404, opened 2022-02-15
Bug:  Performance of Mage Dependent on Installation Style?


sheldonhull:
**Bug Description**
I'm seeing some long-run times in a few environments for a moderate sized monorepo I have. 

**What did you do?**

run `mage`.

**What did you expect to happen?**

Run `mage` and it should be very quick to return list of tasks. 

**What actually happened?**

macOS 

- Using `go install` approach for mage: `mage` takes 18 seconds to return just the help output. 
- I removed this and flipped to the `asdf install mage` version, and it took 3 seconds.
- Debian devcontainer on this mac - takes 3 seconds. 

**Environment**
 - Mage Version: I installed via asdf (latest)
 - OS: macOS, Linux (debian devcontainer)

**Additional context**

Low priority, just wanted to raise this up so I can update it if I run into any further issues on compile time that seem to point towards a specific environment or setup.

natefinch:
The binary is still the same code, the way you install should not matter at all. IIs this something you were able to reproduce, @sheldonhull ?

sheldonhull:
@natefinch Close it for now. If I can reproduce I'll reopen with more detail. Not sure if specific to mage or security scanning software is to blame.

sheldonhull:
I'll add a follow-up that I need to investigate the caching behavior more in the near future since I use Mage so much. Without setting `MAGEFILE_HASHFAST=1` it tends to take a while to run even normal commands despite no changes to the Magefiles. I haven't quantified this, but it's more experience oriented. No need to reopen, but if I find something concrete I'm going to add it here.

natefinch:
My guess is that this is because your magefile imports so much stuff, that it just has a lot to compile.  And unfortunately, the go compiler has to do kind of a lot of work even to tell that it doesn't need to do any more work.

If you run with -debug, it prints out timestamps as it runs, and you can see where the most time is taken. My guess is that it's all the go compiler, figuring out if it needs to recompile or not. It's not as fast as it probably should be.

sheldonhull:
@natefinch if there are no magefile changes, isn't the cached file supposed to be used and operate the same as `MAGEFILE_HASHFAST=1` though? That's the big question I guess. I get that rebuilding `-f` takes 5-10 seconds if it's a big monorepo. However, I'd though the cached output file already being ready to go would be used. I think I need to look into Mage's source so I'm less ignorant on that and might better understand it then.

natefinch:
[...]

PR #440, opened 2022-08-27
refactor(listGoFiles): remove go list dependency

jaredallard:
**What this PR does**: This PR removes the usage of `go list` in favor of parsing the files with the `go/build` package. This speeds up the discovery of go files from ms to us, which combined with `MAGEFILE_HASHFAST` can reduce the time for a re-run of `mage` to ~200ms.

I was getting a little irritated with every run of mage being _at least_ 500-600ms due to `go list` taking a min of `200ms` and being called twice, go list does a lot more then we were using it for. So, I propose we parse it instead.

**Notes for my reviewer**: I did this pretty quickly, and without that much regard for style. I'm happy to split this up and clean it up if we're OK with the approach. Thank you for reviewing :sparkles:

Related to https://github.com/magefile/mage/issues/404

### Performance Measurements

**This PR:**

```bash
$ MAGEFILE_HASHFAST=true hyperfine -i './bin/mage -debug -l'
Benchmark 1: ./bin/mage -debug -l
  Time (mean ± σ):     182.7 ms ±   7.7 ms    [User: 33.5 ms, System: 84.2 ms]
  Range (min … max):   171.9 ms … 199.5 ms    14 runs
```

**1.13.0:**

```bash
$ MAGEFILE_HASHFAST=true hyperfine -i "$HOME/.asdf/installs/mage/1.13.0/bin/mage -debug -l"
Benchmark 1: /Users/jaredallard/.asdf/installs/mage/1.13.0/bin/mage -debug -l
  Time (mean ± σ):     529.2 ms ±  12.7 ms    [User: 196.9 ms, System: 433.4 ms]
  Range (min … max):   507.8 ms … 555.5 ms    10 runs
```

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -476,7 +476,7 @@ func listGoFiles(magePath, goCmd, tag string, envStr []string) ([]string, error)
 	if !filepath.IsAbs(magePath) {
 		cwd, err := os.Getwd()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("can't get current working directory: %v", err)
 		}
 		magePath = filepath.Join(cwd, magePath)
 	}
--- a/mage/main.go
+++ b/mage/main.go
@@ -1,11 +1,13 @@
 package mage
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/sha1"
 	"errors"
 	"flag"
 	"fmt"
+	"go/build/constraint"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


jaredallard:
I probably should've created an issue for this first 😓, please let me know if you'd like me to do that, otherwise I'm happy to chat about this on here or on the Slack!

natefinch:
This sounds pretty great. I'm busy this weekend but I can take a look tomorrow. Thanks a lot for this proposal!

jaredallard:
> This sounds pretty great. I'm busy this weekend but I can take a look tomorrow. Thanks a lot for this proposal!

Sorry to ping, but any update on a review? Totally understand if you're busy!

mage/main.go
```diff
@@ -468,35 +469,54 @@ type mainfileTemplateData struct {
...
-	cmd.Dir = magePath
-	b, err := cmd.Output()
+// listGoFiles returns a list of all .go files in a given directory,
+// matching the provided tag
+func listGoFiles(magePath, goCmd, tag string, envStr []string) ([]string, error) {
+	origMagePath := magePath
+	if !filepath.IsAbs(magePath) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, err
```
natefinch, line 479: everywhere we return an error, we should be prepending some kind of additional info onto it, so that if it bubbles up to the user, they get useful debugging info.

In this case, something like
```suggestion
			return nil, fmt.Errorf("can't get current working directory: %v", err)
```
Assume the level above will add the bit about the fact that it was calling listGoFiles.

(note that we use %v instead of %w, so it'll still compile with older versions of go)

mage/main.go
```diff
@@ -468,35 +469,54 @@ type mainfileTemplateData struct {
...
 	if err != nil {
-		stderr := buf.String()
-		// if the error is "cannot find module", that can mean that there's no
-		// non-mage files, which is fine, so ignore it.
-		if !strings.Contains(stderr, "cannot find module for path") {
-			if tags == "" {
-				return nil, fmt.Errorf("failed to list un-tagged gofiles: %v: %s", err, stderr)
-			}
-			return nil, fmt.Errorf("failed to list gofiles tagged with %q: %v: %s", tags, err, stderr)
+		return nil, err
```
natefinch, line 486:
```suggestion
		return nil, fmt.Errorf("error parsing environment variables: %v", err)
```

natefinch (approved):
Other than a couple very minor additions, this looks great.

mage/main.go
```diff
@@ -468,35 +469,54 @@ type mainfileTemplateData struct {
...
+
+	pkg, err := bctx.Import(".", bctx.Dir, 0)
+	if err != nil {
+		if _, ok := err.(*build.NoGoError); ok {
+			return []string{}, nil
+		}
+
+		// Allow multiple packages in the same directory
+		if _, ok := err.(*build.MultiplePackageError); !ok {
+			return nil, err
```
jaredallard, line 509:
```suggestion
			return nil, fmt.Errorf("failed to parse go source files: %v", err)
```

mirogta (dismissed):
LGTM - with the updated meaningful error messages

jaredallard:
FYI I'm going to be using `mage` a lot more, I'd be happy to help drive some of the improvements here if you all have a priority list of some I could tackle! Thanks for the awesome project! ✨

natefinch (dismissed):
This looks really good.  Thanks so much. Tried it on my local machine and it's definitely faster.

jaredallard:
Thank you so much @natefinch! Is there a timeline I could get for releasing this? Eager to get this through `asdf` 👀

natefinch:
I can make a release tonight:)

PR #441, opened 2022-09-15
feat: adds alias for the kebab cases of the original targets.

### LINKED ISSUE #311: Better support for kebab-case targets
I haven't seen this discussed before, so I wanted to open a discussion.

In my experience, most Makefiles that are not just cramming words together (e.g. `installhtml`, `installps`) use kebab-cased targets (e.g. `install-html`, `install-ps`) instead of camelCased targets (e.g. `installHtml`, `installPs`).  It would be helpful in migrating existing Makefiles and teams to mage if mage could more easily accommodate this pattern.

I'm aware that aliases can help here, but mage Help output only shows the primary targets after running them through the `lowerFirst` template func.

Two ideas:

1. Add an option to show an alias (if one exists) instead of the primary, lowerFirst target in the Help output.  The user could just setup kebab-cased aliases for multi-word targets and the Help output would play along.
1. Add an option to use kebab-casing by default instead of lowerFirst.  This would avoid the need for manual aliases from option 1.

jcchavezs:
Most of the experience in make is around chaining targets using kebab case e.g. make build test e2e-test which isn't supported by mage. This change automatically adds alias for the kebabcase version of the targets if they don't exist.

Closes #311

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -589,6 +589,38 @@ func Compile(goos, goarch, ldflags, magePath, goCmd, compileTo string, gofiles [
 	return nil
 }
 
+var (
+	matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+	matchAllCap   = regexp.MustCompile("([a-z0-9])([A-Z])")
+)
+
+// camelToKebab turns a camelcase string into a kebab case one.
+func camelToKebab(s string) string {
+	ks := matchFirstCap.ReplaceAllString(s, "${1}-${2}")
+	ks = matchAllCap.ReplaceAllString(ks, "${1}-${2}")
+	return strings.ToLower(ks)
+}
+
+// backfillKebabAliases backfill the existing alias with those infered from camelcase targets.
+func backfillKebabAliases(aliases map[string]*parse.Function, funcs []*parse.Function) map[string]*parse.Function {
+	newAliases := aliases
+	if newAliases == nil {
+		newAliases = make(map[string]*parse.Function)
+	}
+	for _, f := range funcs {
+		kebabAlias := camelToKebab(f.Name)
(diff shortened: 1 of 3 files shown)


natefinch:
The problem with this is that the mage list output text can only really show one form.  I don’t really want to support configurations to generate different names, since then we get into configuration hell. 

I would be in favor of showing aliases in the help docs (I can’t remember if they get shown in the full command docs (i.e. mage -h <cmd> but if they’re not, then I think that would be good.  

I don’t like replacing the default generated command name with the alias in the help docs, because an alias is, by definition, an alternate name.  it’s a way to make a script still work with the old name. 

And I’m not a huge fan of generating kebab case targets automatically, because that’s a fairly significant change in behavior that could have unintended side effects on other people’s magefiles. It also really only applies to a very small use case, and has a straightforward work around (i.e. add the aliases manually).

jcchavezs:
Thanks for the feedback @natefinch!

So to be clear, are you fine with autogenerated aliases (notice we don't override user defined aliases if they exist under the kebab-case name) and display the aliases in the help command? I think that is fine, I wouldn't replace the original commands as magic is always confusing.

mage/main.go
```diff
@@ -609,6 +609,38 @@ func Compile(goos, goarch, ldflags, magePath, goCmd, compileTo string, gofiles [
...
+	if newAliases == nil {
+		newAliases = make(map[string]*parse.Function)
+	}
+	for _, f := range funcs {
+		kebabAlias := camelToKebab(f.Name)
+		if kebabAlias == f.Name {
+			continue
+		}
+
+		if _, ok := newAliases[f.Name]; !ok {
```
jcchavezs, line 636: We don't replace existing aliases.

PR #456, opened 2023-02-16
Output Aliases when running `mage -l`

Daegalus:
We use mage for a few projects, and noticed a minor discovery problem of the aliases we setup for our targets. While yes you can run `mage -h target` to see aliases, that's not an obvious command to know about.

I have added the output of Aliases and to what targetname they point to in the `mage -l` output. That way it is more easily discoverable.

If there are no aliases setup, the Aliases section won't show up. I used an actual go if block instead of as a template one as it made `err` variable management a bit simpler. But I can make adjustments if it is preferred.

No new tests were added as a bunch of existing tests would fail if `Aliases` always showed up, and the ImportLists test fails if it doesn't show up.

I can add additional tests if requested though.

### DIFF
--- a/mage/import_test.go
+++ b/mage/import_test.go
@@ -28,6 +28,9 @@ Targets:
   zz:buildSubdir2    Builds stuff.
   zz:ns:deploy2*     deploys stuff.
 
+Aliases:
+  nsd2    -> zz:ns:deploy2
+
 * default target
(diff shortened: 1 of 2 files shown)


daegalus:
Its been 2 years, and I wouldn't mind knowing if this will ever be looked at or considered?

PR #474, opened 2023-06-25
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.

This patch was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


natefinch:
I like this idea, but passing in a giant opaque number is kind of a horrible UX. Also, all the other environment variables are prefixed with MAGEFILE, so it would at least need to follow that pattern.

I'd be in favor of having MAGEFILE_SOURCE_DATE, which takes an RFC3339 timestamp e.g. "2023-12-20T13:59:59Z" or "2023-12-20T13:59:59-05:00". Then at least the timestamp is human readable and writeable, unlike unix epoch.

bmwiedemann:
We could, though that would mean that we need some place where we convert `SOURCE_DATE_EPOCH` (which we set already anyway for the 100 other tools that use it) into `MAGEFILE_SOURCE_DATE`.

It would be an improvement over the status quo in either case.

Issue #483, opened 2023-10-10
Documentation for MAGEFILE_CACHE is not consistent with actual behavior on Windows


WilliamRoyNelson:
https://github.com/magefile/mage/blob/master/site/content/environment/_index.en.md?plain=1#L14-L17
```
## MAGEFILE_CACHE

Sets the directory where mage will store binaries compiled from magefiles
(default is $HOME/.magefile)
```

https://github.com/magefile/mage/blob/9f54e0f83e2a8d2976c07037ad74aa20c62797a5/site/content/howitworks/_index.en.md?plain=1#L15-L18
```
## Binary Cache

Compiled magefile binaries are stored in $HOME/.magefile.  This location can be
customized by setting the MAGEFILE_CACHE environment variable.
```

https://github.com/magefile/mage/blob/9f54e0f83e2a8d2976c07037ad74aa20c62797a5/mg/runtime.go#L110-L114
```
	switch runtime.GOOS {
	case "windows":
		return filepath.Join(os.Getenv("HOMEDRIVE"), os.Getenv("HOMEPATH"), "magefile")
	default:
		return filepath.Join(os.Getenv("HOME"), ".magefile")
```

On Windows, instead of creating a folder at `%USERPROFILE%\.magefile` as would be expected by the documentation, a folder is created at `%USERPROFILE%\magefile`

WilliamRoyNelson:
Docs fix PR: https://github.com/magefile/mage/pull/484

mloskot:
This can be closed since 
- #484

has been merged by @natefinch 

@WilliamRoyNelson It is a good idea to use the GitHub features of [Closing Issues via Commit Messages](https://github.blog/2013-01-22-closing-issues-via-commit-messages/) next time, so related issues are not left dangling after merge.

WilliamRoyNelson:
> This can be closed since 
> - #484
> 
> has been merged by @natefinch 
> 
> @WilliamRoyNelson It is a good idea to use the GitHub features of [Closing Issues via Commit Messages](https://github.blog/2013-01-22-closing-issues-via-commit-messages/) next time, so related issues are not left dangling after merge.

That is an awesome feature that I was not aware of. I will make sure to do that in the future!

PR #484, opened 2023-10-10
Update documentation to include Windows Magefile Cache

WilliamRoyNelson:
Fix for documentation issue: https://github.com/magefile/mage/issues/483
Alternately, the path could be changed to `%USERPROFILE%\.magefile` for Windows, but that could cause issues when upgrading.


natefinch (approved):
Thanks! I really appreciate the update to the docs :)
I intentionally didn't use .magefile in windows, because it can be a lot harder to "find" directories that start with a period in windows, and it seems a lot less common to use that style of directory on that OS.

PR #490, opened 2023-11-18
Correct test input code to support older go versions

perrito666:
Samples used for testing need to have build tags for new and old style:

```
//go:build tag
// +build tag
```

The particular problem here, seems to be that 1.16 was actually importing the package referred in the `TestGoModules` test (`import "golang.org/x/text/unicode/norm"`) which, in the CI system's version lacks an old style go tag comment. Hence go complaining of a `//go:build` without a `// +build`. I made all files bi-tagged for consistency.

### DIFF
--- a/bootstrap.go
+++ b/bootstrap.go
@@ -1,4 +1,5 @@
-//+build ignore
+//go:build ignore
+// +build ignore
 
 package main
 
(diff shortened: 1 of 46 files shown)


natefinch (approved):
Thank you!  Nothing should be imported outside the stdlib anyway. I don't want this tool to require outside imports when you build it.

PR #491, opened 2023-11-18
Draft attempt at extra args

perrito666:
So, this is an idea mostly, i have not been able to test this as running `mage` just gets the `mg` from upstream, perhaps i should not use upstream?

The idea is to support this:

```go
//go:build mage
// +build mage

package main

import (
	"fmt"

	mg "../../go/github.com/magefile/mage/mg"
)

func TargetOne(extra mg.ExtraArgs) error {
	fmt.Printf("%#v\n", extra)
}
```

where ideally extra would become a string slice  `[]string{"-baz", "blah"}` after calling `mage targetone foo bar -- -baz blah`

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -119,6 +119,10 @@ type Invocation struct {
 	GoCmd      string        // the go binary command to run
 	CacheDir   string        // the directory where we should store compiled binaries
 	HashFast   bool          // don't rely on GOCACHE, just hash the magefiles
+
+	// all the arguments that came after --, we do not do any kind of parsing on them as
+	// they are intended for a subprocess and we know nothing about them
+	ExtraArgs []string
 }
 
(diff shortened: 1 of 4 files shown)


perrito666:
Ok @natefinch this works now as I intended, let me know if this is something we want to support.

natefinch:
I think we need a test where we run mage.ParseAndRun() and pass in a raw args list and see that it does the right thing. Right now, the code in mage/main.go in the Parse function is never tested, and that's pretty key.

perrito666:
@natefinch done, added the same test cases, which i consider to be relevant but also using `ParseAndRun`

perrito666:
@natefinch I completely forgot this never got merged :p I think the idea is still valid and might be a good way to implement optional args (adding special cases of the base types)

PR #499, opened 2024-01-20
chore(docs): updated README and site to include installation using go install command

egor-denysenko:
Updated docs and website index with instructions on how to install mage using the `go install` command


nojaf:
Could someone please merge this? I bumped into this problem today.

natefinch (approved):
Thank you!

PR #504, opened 2024-05-01
Use %w to wrap upstream errors

### LINKED ISSUE #503: Bug: Errors aren't created with %w, breaking errors.Is
**Bug Description**

In one of my build targets, I use `sh.Output` to check the output of a command. For example:

```go
output, err := sh.Output("command", "--version")
```

But in my case, it's completely normal for "command" not to exist sometimes, so I want to handle that error differently from other failures that `sh.Output` might return. When the command doesn't exist, the underlying `os.Exec` code will return an `fs.PathError`, and the idiomatic way to check for that is to use `errors.Is`:

```go
if err != nil {
    if !errors.Is(err, fs.ErrNotExist) {
        // This is really an error.
        return err
    }
    // File-doesn't-exist is an expected condition that we can handle.
}
```

But `errors.Is` returns false for errors returned by `sh.Output`.

**What did you do?**

As described above, I tried to use `errors.Is` to inspect errors returned by `sh.Output` and other command-executing functions from `sh`.

**What did you expect to happen?**

I expected `errors.Is` to recognize errors from `sh` functions as any of various common system errors, particularly `fs.ErrNotExist`.

**What actually happened?**

`errors.Is` fails to acknowledge that any `sh` error is like any other known error.

**Environment**

 - Mage Version: v1.15.0
 - OS: Linux

**Additional context**

This happens because errors in `sh` are constructed using `fmt.Errorf("...%v", err)`. That inserts the value of `err.String()` but does not actually _wrap_ the error, so any additional information about `err` is discarded. Changing it to `fmt.Errorf("...%w", err)` would solve it without affecting any other behavior. Error wrapping was [introduced in Go 1.13](https://go.dev/doc/go1.13#error_wrapping).

My workaround for this issue is to inspect the text of the error message instead:

```go
if err != nil {
    if !strings.Contains(err.Error(), "no such file or directory") {
        // This is really an error
        return err
    }
    // File-doesn't-exist is an expected condition that we can handle.
}
```

rkennedy:
Wrapping errors instead of merely embedding error messages allows calling code to use `errors.Is` and `errors.As` to more precisely check the reasons for various errors instead of having to rely only on substring searches.

Fixes #503

Since error-wrapping was only introduced in Go 1.13, accepting this PR would involve dropping support for 1.11 and 1.12 (and even earlier, if the readme's claim of supporting all the way back to 1.7 is still accurate). I realize that may jeopardize this PR because it's definitely not a change to make lightly.


TotallyGamerJet:
Why not just implement a wrapped error type specifically for mage? That way version compatibility can be maintained

rkennedy:
> Why not just implement a wrapped error type specifically for mage? That way version compatibility can be maintained

Well, because my initial reaction was that "just" was doing quite a bit of work in that question! But it turned out not to be so bad after all. I didn't have to do anything crazy like re-implement `errors.Is`, just write a wrapper that implements `Error()` and `Unwrap()`, and then add a bunch of calls to use the new error type. The call sites look cumbersome _within this library_, but _consumers_ of the library should never notice a difference. Using build tags, we can ensure that the new functionality is tested on versions that support it.

Thanks for the nudge toward a better solution.

TotallyGamerJet:
Do we want to actually export the `WrapError` function for anyone to use? If not, I'd move it into the `internal/` package. 

Also, why not make the function signature like this? It improves the calling site.

```go
 WrapErrorF(err, "error running \"%s %s\": %v\n%s", cmd, strings.Join(args, " "), err, errMsg)

func WrapErrorF(underlying error, format string, args ...interface{}) error {
	return &wrappedError{
		underlyingError: underlying,
		stringError:     fmt.Errorf(format, args...),
	}
}
```

PR #513, opened 2024-10-16
Traverse upwards to find ./magefiles directory

morganhein:
This is a follow up from the other PR that has stalled.

I am using mage in a monorepo, with many layers of folders that may contain magefiles. It's been a major improvement working with this repo when mage walks up the tree to find the closest magefiles folder.

This PR causes the following affects:
1. By default, it searches for a `magefiles` folder within the current directory
2. If not found, it begins walking up the tree to find the folder
3. If a `.git` folder is detected, it stops. This is based on the assumption that many magefiles probably should be scope to their git repository
4. If the root is encountered, it also stops.
5. You can disable this behavior by passing -no-traverse or setting `MAGEFILE_NO_TRAVERSE` as an environment variable

I can see an argument for some sort of `.magebuild` or `.mageroot` file that also can be used to indicate traversal stopping.

I did not update the website. I'm open to that, however wanted to get some opinions on this before I work on it much further.

Thanks!

### DIFF
--- a/mage/main.go
+++ b/mage/main.go
@@ -99,26 +99,27 @@ func Main() int {
 
 // Invocation contains the args for invoking a run of Mage.
 type Invocation struct {
-	Debug      bool          // turn on debug messages
-	Dir        string        // directory to read magefiles from
-	WorkDir    string        // directory where magefiles will run
-	Force      bool          // forces recreation of the compiled binary
-	Verbose    bool          // tells the magefile to print out log statements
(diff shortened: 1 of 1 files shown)


DavidGamba:
In my implementation of this for a similar tool I have a case for when the current folder is the `magefiles` folder. I suppose you just walk up and find it but I though of it as an edge case that returns '.' rather than doing some work to make it relative.

https://github.com/DavidGamba/dgtools/blob/master/bake/build.go#L64

PR #526, opened 2025-06-19
deprecate sha1

miparnisari:
See https://pkg.go.dev/crypto/sha1

> Package sha1 implements the SHA-1 hash algorithm as defined in [RFC 3174](https://rfc-editor.org/rfc/rfc3174.html).

> SHA-1 is cryptographically broken and should not be used for secure applications.


miparnisari:
Hi @natefinch and @perrito666, could you review this? Thank you 😄

PR #536, opened 2026-02-06
feat: Add ability to run commands at given path

AtzeDeVries:
Sometimes a command needs to run at a given path. The current `sh` package does not have this ability. This PR adds the feature. It adds "At" to all the run functions where you can supply it with a path were it needs to run. The functions are added under a new name so this change is backwards compatable.

### DIFF
--- a/sh/cmd.go
+++ b/sh/cmd.go
@@ -212,7 +212,7 @@ func run(env map[string]string, stdout, stderr io.Writer, pwd string, cmd string
 	}
 	// To protect against logging from doing exec in global variables
 	if mg.Verbose() {
-		log.Println("exec:", cmd, strings.Join(quoted, " "))
+		log.Printf("exec: %s %s, pwd: %s\n", cmd, strings.Join(quoted, " "), pwd)
 	}
 	err = c.Run()
 	return CmdRan(err), ExitStatus(err), err


sh/cmd.go
```diff
@@ -101,7 +167,7 @@ func OutputWith(env map[string]string, cmd string, args ...string) (string, erro
 // Ran reports if the command ran (rather than was not found or not executable).
 // Code reports the exit code the command returned if it ran. If err == nil, ran
 // is always true and code is always 0.
-func Exec(env map[string]string, stdout, stderr io.Writer, cmd string, args ...string) (ran bool, err error) {
+func Exec(env map[string]string, stdout, stderr io.Writer, pwd string, cmd string, args ...string) (ran bool, err error) {
```
michel-laterman: I'm another user who is interested in this feature; however, I think that adding a new required parameter to the function signature would be a breaking change to the API, can you instead introduce something like:

```go
func ExecAt(env map[string]string, stdout, stderr io.Writer, pwd string, cmd string, args ...string) (ran bool, err error)
```

and change `Exec` so it just does:
```go
func Exec(env map[string]string, stdout, stderr io.Writer, cmd string, args ...string) (ran bool, err error) {
	ExecAt(env, stdout, stderr, "", cmd, args...)
}
```

sh/cmd_test.go
```diff
@@ -68,5 +81,46 @@ func TestAutoExpand(t *testing.T) {
...
+
+func TestSettingPwd(t *testing.T) {
+	pwd := "/"
+	out := &bytes.Buffer{}
+	ran, err := ExecAt(nil, out, nil, pwd, "pwd")
+	if err != nil {
+		t.Fatalf("unexpected error from runner: %#v", err)
+	}
+	if !ran {
+		t.Errorf("expected ran to be true but was false.")
```
michel-laterman:
```suggestion
		t.Error("expected ran to be true but was false.")
```

sh/cmd_test.go
```diff
@@ -68,5 +81,46 @@ func TestAutoExpand(t *testing.T) {
...
+	}
+	if out.String() != fmt.Sprintf("%s\n", pwd) {
+		t.Errorf("expected %s, got %q", fmt.Sprintf("%s\n", pwd), out)
```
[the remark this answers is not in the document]
michel-laterman: Why not use `t.Errorf("expected %s, got %q", pwd, out)` or `t.Errorf("expected %s\n, got %q", pwd, out)` (if a newline is needed)?

PR #551, opened 2026-04-16
Support Tab Completion

natefinch:
Tab completion for zsh, bash, fish, and powershell installed with `mage -install <shellname>`.

Add hidden `-autocomplete` to return just the names of targets so you can use your own autocomplete if you don't like Mage's.

Also while we're at it, made `mage -l` lightning fast since it no longer compiles the full code, since it already knows everything it needs to output the list of targets before compile.

### DIFF
--- a/mage/args_test.go
+++ b/mage/args_test.go
@@ -207,7 +207,6 @@ func TestOptionalArgsOmitted(t *testing.T) {
 		Dir:    "./testdata/optargs",
 		Stderr: stderr,
 		Stdout: stdout,
-		Keep:   true,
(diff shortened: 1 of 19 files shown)


natefinch:
@copilot fix the linter errors.

mage/args_test.go
```diff
@@ -207,6 +207,7 @@ func TestOptionalArgsOmitted(t *testing.T) {
 		Dir:    "./testdata/optargs",
 		Stderr: stderr,
 		Stdout: stdout,
+		Keep:   true,
```
natefinch:
```suggestion
```

